PR #4, opened 2016-07-27
Add an error message when the reason for exec failing is that JAVA_HO…

jamesneale:
…ME is incorrect


jamesneale:
Probably are much better ways to go about this.

uschi2000:
want to add a test?

uschi2000:
actually, nevermind.

PR #6, opened 2016-08-02
Run ptimports on all files

nmiyake:


uschi2000:
Nice. How can I do this in an automated fashion?

nmiyake:
Hope to do so by hooking up gödel (makes it possible to run on all files easily, do this as a check in `circle.yml`, and install a git commit hook to enforce locally as well). Main impediment is just open-sourcing so it's accessible on public GH.com -- definitely hope for this to happen, but think we want to play-test internally for a bit first.

uschi2000:
Cool, cheers Nick.

PR #10, opened 2016-10-25
Check Java binary command for safety

uschi2000:
This change is [Reviewable](https://reviewable.io/reviews/palantir/go-java-launcher/10)

### DIFF
--- a/launchlib/launcher.go
+++ b/launchlib/launcher.go
@@ -21,12 +21,25 @@ import (
 	"path"
 	"strings"
 	"syscall"
+	"regexp"
 )
 
 const (
-	TemplateDelimsOpen  = "{{"
+	TemplateDelimsOpen = "{{"
 	TemplateDelimsClose = "}}"
+	// Matches characters disallowed in paths we allow to be passed to exec()
+	ExecPathBlackListRegex = `[^\w.\/_\-]`
 )
+
+// Returns true iff the given path is safe to be passed to exec(): must not contain funky characters and be a valid file
+func verifyPathIsSafeForExec(execPath string) string {
+	unsafe, _ := regexp.MatchString(ExecPathBlackListRegex, execPath)
+	_, statError := os.Stat(execPath)
+	if (unsafe || statError != nil) {
+		panic("Failed to determine is path is safe to execute: " + execPath)
+	}
+	return execPath
+}
 
 type processExecutor interface {
 	Exec(executable string, args []string, env []string) error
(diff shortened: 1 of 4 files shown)


bmoylan:
:lgtm: you could use the `os.FileInfo` for more checks if you want, but the sanitization is the most important part

---

Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

_[launchlib/launcher.go, line 37 at r1](https://reviewable.io:443/reviews/palantir/go-java-launcher/10#-KUtUAjgKfmHsQ3GOBn3:-KUtUAjgKfmHsQ3GOBn4:bpl1sit) ([raw file](https://github.com/palantir/go-java-launcher/blob/8f398371320f95640c3429dd53d7a27e6b77d48d/launchlib/launcher.go#L37)):_

> ``` Go
> func verifyPathIsSafeForExec(execPath string) string {
>   unsafe, _ := regexp.MatchString(ExecPathBlackListRegex, execPath)
>   _, statError := os.Stat(execPath)
> ```

Do you want to use the `FileInfo` returned from `Stat` for anything? You can check that it's a regular executable file, for instance.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/go-java-launcher/10#-:-KUtUcoPZxp_cQdWJdjy:bypo6cm)_

launchlib/launcher.go
```diff
@@ -21,13 +21,26 @@ import (
...
 	"syscall"
+	"regexp"
 )
 
 const (
-	TemplateDelimsOpen  = "{{"
+	TemplateDelimsOpen = "{{"
 	TemplateDelimsClose = "}}"
+	// Matches characters disallowed in paths we allow to be passed to exec()
+	ExecPathBlackListRegex = `[^\w.\/_\-]`
```
mglazer, line 31: Some services include '~' in the path name. This seems to fail for that.

launchlib/launcher.go
```diff
@@ -21,13 +21,26 @@ import (
...
 )
 
+// Returns true iff the given path is safe to be passed to exec(): must not contain funky characters and be a valid file
+func verifyPathIsSafeForExec(execPath string) string {
+	unsafe, _ := regexp.MatchString(ExecPathBlackListRegex, execPath)
+	_, statError := os.Stat(execPath)
+	if (unsafe || statError != nil) {
```
mglazer, line 38: Agree that you might want to check that the exec bit is set here. Might also want to check that it's readable.

launchlib/launcher.go
uschi2000: I don't think we expand `~` to the home directory (this is a shell feature, right), so this won't have the intuitive effect anyway?

launchlib/launcher.go
uschi2000: I will leave that for another PR if that's OK?

PR #15, opened 2016-12-05
Use gödel to build and publish product

nmiyake:
This change is [Reviewable](https://reviewable.io/reviews/palantir/go-java-launcher/15)


uschi2000:
The circle build passes although a test fails.

uschi2000:
Where did the imports of the form `go get -u -v golang.org/x/tools/go/gcimporter15` go?

nmiyake:
Yes the Circle thing I'm looking at now. It's actually an interesting case we've seen elsewhere, where `go test` doesn't report the failure, but the JUnit XML output does. The XML is actually correct -- right now the "main" test isn't working like it's supposed to -- since the function calls `exec`, that kills the test process so only 1 of the 3 tests in that suite run. `go test` doesn't recognize this as a failure, but the JUnit output does, and that's why the report has a failure.

Working on fixing the test more formally by making them integration tests (gödel has first-class support for doing this, and it makes sense in this case).

As for the imports, I'm guessing those were used by one of the other tools that was being fetched by `go get` since the `gcimporter15` package isn't used by any of the core code in this project. Pretty much all of that code is included as part of the gödel executable.

uschi2000:
Cool, thanks

nmiyake:
OK, this should be ready for review now.

nmiyake:
Also, ran the `publish` task manually locally and verified that it properly creates and uploads the product to Bintray in the same way that the previous publish script did.

uschi2000:
OK, let's try this.

nmiyake:
Great, looks like merge build completed successfully. Do you want to try to tag a minor version bump to ensure that publishing works as well, or would you rather wait for a real release?

uschi2000:
Tagged 1.1.1-rc1

nmiyake:
Great, looks like publish succeeded!

Is there an easy way to validate that the artifact behaves as intended? If not I think we're still fine, but if there is might be worth trying it out just to verify that everything is okay. Otherwise, I think we're set!

uschi2000:
Tried in GJD, looks good.

uschi2000:
Thanks, Nick!

nmiyake:
Great to hear -- np!

PR #18, opened 2016-12-06
Update gödel to 0.11.1

nmiyake:
Includes fix that allows the downloading of the distribution to
fall back to using curl if downloading using wget fails.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/go-java-launcher/18)


nmiyake:
@uschi2000 this fixes issues that some people have seen with downloading the distribution.

uschi2000 (approved):

PR #19, opened 2016-12-08
Add support for arbitrary execution

### LINKED ISSUE #20: support arbitrary process execution
I want to use this go-launcher framework to securely launch non-java services (such as postgres/influx/grafana) that depend on setting environmental params.

tomkozlowski:
Also... rename to `go-launcher`

Also includes a breaking change in schema version, but given that I'm also proposing renaming the repo, that seems reasonable

Fixes #20 

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/go-java-launcher/19)

### DIFF
--- a/main.go
+++ b/main.go
@@ -16,27 +16,14 @@ package main
 
 import (
 	"fmt"
-	"io/ioutil"
 	"os"
 
 	"github.com/palantir/go-java-launcher/launchlib"
 )
 
-func LaunchWithConfig(staticConfigFile, customConfigFile string) {
-	staticData, err := ioutil.ReadFile(staticConfigFile)
-	if err != nil {
-		panic("Failed to read static config file: " + staticConfigFile)
-	}
-	staticConfig := launchlib.ParseStaticConfig(staticData)
-
-	var customConfig launchlib.CustomLauncherConfig
-	if customData, err := ioutil.ReadFile(customConfigFile); err != nil {
-		fmt.Println("Failed to read custom config file, assuming no custom config:", customConfigFile)
-	} else {
-		customConfig = launchlib.ParseCustomConfig(customData)
-	}
-
-	launchlib.Launch(&staticConfig, &customConfig)
+func Exit1WithMessage(message string) {
+	fmt.Fprintln(os.Stderr, message)
+	os.Exit((1))
 }
 
 func main() {
@@ -45,7 +32,7 @@ func main() {
 
 	switch numArgs := len(os.Args); {
 	case numArgs > 3:
-		panic("Usage: javalauncher [<path to StaticLauncherConfig> [<path to CustomLauncherConfig>]]")
+		Exit1WithMessage("Usage: go-java-launcher <path to StaticLauncherConfig> [<path to CustomLauncherConfig>]")
 	case numArgs == 2:
 		staticConfigFile = os.Args[1]
 	case numArgs == 3:
@@ -53,5 +40,8 @@ func main() {
 		customConfigFile = os.Args[2]
 	}
 
-	LaunchWithConfig(staticConfigFile, customConfigFile)
+	launchErr := launchlib.LaunchWithConfig(staticConfigFile, customConfigFile)
+	if launchErr != nil {
+		Exit1WithMessage(launchErr.Error())
+	}
 }
--- a/README.md
+++ b/README.md
@@ -3,45 +3,63 @@
 
 # go-java-launcher
 
-A simple Go program for launching Java programs from a fixed configuration. This program replaces Gradle-generated Bash
+A simple Go program for launching programs from a fixed configuration. This program replaces Gradle-generated Bash
 launch scripts which are susceptible to attacks via injection of environment variables of the form `JAVA_OPTS='$(rm -rf
 /)'`.
 
 The launcher accepts as configuration two YAML files as follows:
 
 ```yaml
-# StaticLauncherConfig
-# The type of configuration, must be the string "java"
+# StaticLauncherConfig - java version
+# REQUIRED - The type of configuration, must be the string "java"
 configType: java
-# The version of the configuration format, must be the integer 1
+# REQUIRED - The version of the configuration format, must be the integer 1
 configVersion: 1
-# The main class to be run
+# REQUIRED - The main class to be run
 mainClass: my.package.Main
-# Path to the JRE, defaults to the JAVA_HOME environment variable if unset
+# OPTIONAL - Path to the JRE, defaults to the JAVA_HOME environment variable if unset
 javaHome: javaHome
-# The classpath entries; the final classpath is the ':'-concatenated list in the given order
+# REQUIRED - The classpath entries; the final classpath is the ':'-concatenated list in the given order
 classpath:
   - ./foo.jar
-# Environment Variables to be set in the environment
+# OPTIONAL - Environment Variables to be set in the environment (Note: cannot be referenced on args list)
 env:
   CUSTOM_VAR: CUSTOM_VALUE
-# JVM options to be passed to the java command
+# OPTIONAL - JVM options to be passed to the java command
 jvmOpts:
   - '-Xmx1g'
-# Arguments passed to the main method of the main class
+# OPTIONAL - Arguments passed to the main method of the main class
+args:
+  - arg1
+```
+
+```yaml
+# StaticLauncherConfig - executable version
+# REQUIRED - The type of configuration, must be the string "executable"
+configType: java
+# REQUIRED - The version of the configuration format, must be the integer 1
+configVersion: 1
+# OPTIONAL - Environment Variables to be set in the environment (Note: cannot be referenced on args list)
+env:
+  CUSTOM_VAR: CUSTOM_VALUE
+# REQUIRED - The full path to the executable file, limited to whitelisted values (java, postgres, influxd, grafana-server)
+executable: "{{CWD}}/service/bin/postgres"
+# OPTIONAL - Arguments passed to the main method of the excutable or main class
 args:
   - arg1
 ```
 
 ```yaml
 # CustomLauncherConfig
+# REQUIRED - The type of configuration, must be the string "java" or "executable"
 configType: java
+# REQUIRED - The version of the configuration format, must be the integer 1
 configVersion: 1
-# Environment variables to be set in the runtime environment
+# OPTIONAL - Environment Variables to be set in the environment, will override defaults in static config (Note: cannot be referenced on args list)
 env:
   CUSTOM_VAR: CUSTOM_VALUE
   CUSTOM_PATH: '{{CWD}}/some/path'
-# JVM options to be passed to the java command
+# Additional JVM options to be passed to the java command, will override defaults in static config. Ignored if configType is "executable"
 jvmOpts:
   - '-Xmx2g'
 ```
--- a/integration_test/integration_test.go
+++ b/integration_test/integration_test.go
@@ -28,14 +28,14 @@ func TestMainMethod(t *testing.T) {
 	require.NoError(t, err, "failed: %s", output)
 
 	// part of expected output from launcher
-	assert.Regexp(t, `Argument list to Java binary: \[.+/bin/java -Xmx4M -Xmx1g -classpath .+/github.com/palantir/go-java-launcher/integration_test/test_resources Main arg1\]`, output)
+	assert.Regexp(t, `Argument list to executable binary: \[.+/bin/java -Xmx4M -Xmx1g -classpath .+/github.com/palantir/go-java-launcher/integration_test/test_resources Main arg1\]`, output)
 	// expected output of Java program
 	assert.Regexp(t, `\nmain method\n`, string(output))
 }
 
 func TestPanicsWhenJavaHomeIsNotAFile(t *testing.T) {
 	_, err := runMainWithArgs(t, "test_resources/launcher-static-bad-java-home.yml", "foo")
-	require.Error(t, err, "panic: Failed to determine is path is safe to execute: /foo/bar/bin/java")
+	require.Error(t, err, "error: Failed to determine is path is safe to execute: /foo/bar/bin/java")
 }
 
 func TestMainMethodWithoutCustomConfig(t *testing.T) {
@@ -44,7 +44,7 @@ func TestMainMethodWithoutCustomConfig(t *testing.T) {
 
 	// part of expected output from launcher
 	assert.Regexp(t, `Failed to read custom config file, assuming no custom config: foo`, output)
-	assert.Regexp(t, `Argument list to Java binary: \[.+/bin/java -Xmx4M -classpath .+/github.com/palantir/go-java-launcher/integration_test/test_resources Main arg1\]`, output)
+	assert.Regexp(t, `Argument list to executable binary: \[.+/bin/java -Xmx4M -classpath .+/github.com/palantir/go-java-launcher/integration_test/test_resources Main arg1\]`, output)
 	// expected output of Java program
 	assert.Regexp(t, `\nmain method\n`, string(output))
 }
--- a/launchlib/config.go
+++ b/launchlib/config.go
@@ -16,32 +16,30 @@ package launchlib
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"strconv"
+	"strings"
 
+	"github.com/pkg/errors"
+	"gopkg.in/validator.v2"
 	"gopkg.in/yaml.v2"
 )
 
-type LauncherConfig struct {
-	ConfigType    string `yaml:"configType"`
-	ConfigVersion int    `yaml:"configVersion"`
-}
-
-func (config *LauncherConfig) validate() {
-	if config.ConfigType != "java" {
-		panic(fmt.Sprintf("Can handle configType=java only, found %s", config.ConfigType))
-	}
-	if config.ConfigVersion != 1 {
-		panic(fmt.Sprintf("Can handle configVersion=1 only, found %d", config.ConfigVersion))
-	}
+type JavaConfig struct {
+	JavaHome  string   `yaml:"javaHome"`
+	MainClass string   `yaml:"mainClass" validate:"nonzero"`
+	JvmOpts   []string `yaml:"jvmOpts"`
+	Classpath []string `yaml:"classpath" validate:"nonzero"`
 }
 
 type StaticLauncherConfig struct {
 	LauncherConfig `yaml:",inline"`
+	JavaConfig     `yaml:",inline"`
 	ServiceName    string            `yaml:"serviceName"`
-	MainClass      string            `yaml:"mainClass"`
-	JavaHome       string            `yaml:"javaHome"`
 	Env            map[string]string `yaml:"env"`
-	Classpath      []string          `yaml:"classpath"`
-	JvmOpts        []string          `yaml:"jvmOpts"`
+	Executable     string            `yaml:"executable,omitempty"`
 	Args           []string          `yaml:"args"`
 }
 
@@ -51,25 +49,116 @@ type CustomLauncherConfig struct {
 	Env            map[string]string `yaml:"env"`
 }
 
-func ParseStaticConfig(yamlString []byte) StaticLauncherConfig {
+type LauncherConfig struct {
+	ConfigType    string `yaml:"configType"`
+	ConfigVersion int    `yaml:"configVersion"`
+}
+
+type AllowedLauncherConfigValues struct {
+	ConfigTypes    map[string]struct{}
+	ConfigVersions map[int]struct{}
+	Executables    map[string]struct{}
+}
+
+var allowedLauncherConfigs = AllowedLauncherConfigValues{
+	ConfigTypes:    map[string]struct{}{"java": {}, "executable": {}},
+	ConfigVersions: map[int]struct{}{1: {}},
+	Executables:    map[string]struct{}{"java": {}, "postgres": {}, "influxd": {}, "grafana-server": {}},
+}
+
+func ParseStaticConfig(yamlString []byte) (StaticLauncherConfig, error) {
 	var config StaticLauncherConfig
 	if err := yaml.Unmarshal(yamlString, &config); err != nil {
-		unmarshalErrPanic("StaticLauncherConfig", err)
+		return StaticLauncherConfig{},
+			errors.Wrap(err, "Failed to deserialize Static Launcher Config, please check the syntax of your configuration file")
 	}
-	config.LauncherConfig.validate()
-	return config
+
+	if err := config.LauncherConfig.validateLauncherConfig(); err != nil {
+		return StaticLauncherConfig{}, err
+	}
+
+	if config.ConfigType == "java" {
+		config.Executable = "java"
+		if err := validator.Validate(config.JavaConfig); err != nil {
+			return StaticLauncherConfig{}, err
+		}
+	}
+
+	if err := validateExecutableConfig(config.Executable); err != nil {
+		return StaticLauncherConfig{}, err
+	}
+	return config, nil
 }
 
-func ParseCustomConfig(yamlString []byte) CustomLauncherConfig {
+func GetStaticConfigFromFile(staticConfigFile string) (StaticLauncherConfig, error) {
+	if staticData, err := ioutil.ReadFile(staticConfigFile); err != nil {
+		return StaticLauncherConfig{}, errors.Wrap(err, "Failed to read static config file: "+staticConfigFile)
+	} else if staticConfig, err := ParseStaticConfig(staticData); err != nil {
+		return StaticLauncherConfig{}, err
+	} else {
+		return staticConfig, nil
+	}
+
+}
+
+func ParseCustomConfig(yamlString []byte) (CustomLauncherConfig, error) {
 	var config CustomLauncherConfig
 	if err := yaml.Unmarshal(yamlString, &config); err != nil {
-		unmarshalErrPanic("CustomLauncherConfig", err)
+		return CustomLauncherConfig{},
+			errors.Wrap(err, "Failed to deserialize Custom Launcher Config, please check the syntax of your configuration file")
 	}
-	config.LauncherConfig.validate()
-	return config
+	if err := config.LauncherConfig.validateLauncherConfig(); err != nil {
+		return CustomLauncherConfig{}, err
+	}
+	return config, nil
 }
 
-func unmarshalErrPanic(structName string, err error) {
-	fmt.Printf("Failed to deserialize %s, please check the syntax of your configuration file\n", structName)
-	panic(err)
+func GetCustomConfigFromFile(customConfigFile string) (CustomLauncherConfig, error) {
+	if customData, err := ioutil.ReadFile(customConfigFile); err != nil {
+		fmt.Fprintln(os.Stderr, "Failed to read custom config file, assuming no custom config:", customConfigFile)
+		return CustomLauncherConfig{}, nil
+	} else if customConfig, err := ParseCustomConfig(customData); err != nil {
+		return CustomLauncherConfig{}, err
+	} else {
+		return customConfig, nil
+	}
+}
+
+func (config *LauncherConfig) validateLauncherConfig() error {
+	if _, ok := allowedLauncherConfigs.ConfigTypes[config.ConfigType]; !ok {
+		return fmt.Errorf("Can handle configType=%v only, found %s",
+			toString(allowedLauncherConfigs.ConfigTypes), config.ConfigType)
+	}
+	if _, ok := allowedLauncherConfigs.ConfigVersions[config.ConfigVersion]; !ok {
+		return fmt.Errorf("Can handle configVersion=%v only, found %d",
+			toString(convertMap(allowedLauncherConfigs.ConfigVersions)), config.ConfigVersion)
+	}
+	return nil
+}
+
+func validateExecutableConfig(executable string) error {
+	if executable == "" {
+		return errors.New("Config type \"executable\" requires top-level \"executable:\" value")
+	}
+	if _, ok := allowedLauncherConfigs.Executables[path.Base(executable)]; !ok {
+		return fmt.Errorf("Can handle executable=%v only, found %v",
+			toString(allowedLauncherConfigs.Executables), executable)
+	}
+	return nil
+}
+
+func toString(inputMap map[string]struct{}) string {
+	collection := make([]string, 0, len(inputMap))
+	for k := range inputMap {
+		collection = append(collection, k)
+	}
+	return "{" + strings.Join(collection, ", ") + "}"
+}
+
+func convertMap(intMap map[int]struct{}) map[string]struct{} {
+	stringMap := make(map[string]struct{})
+	for k, v := range intMap {
+		stringMap[strconv.Itoa(k)] = v
+	}
+	return stringMap
 }
--- a/launchlib/config_test.go
+++ b/launchlib/config_test.go
@@ -21,7 +21,14 @@ import (
 )
 
 func TestParseStaticConfig(t *testing.T) {
-	var data = []byte(`
+	for i, currCase := range []struct {
+		name string
+		data string
+		want StaticLauncherConfig
+	}{
+		{
+			name: "java static config",
+			data: `
 configType: java
 configVersion: 1
 mainClass: mainClass
@@ -38,25 +45,56 @@ jvmOpts:
 args:
   - arg1
   - arg2
-`)
-	want := StaticLauncherConfig{
-		LauncherConfig: LauncherConfig{
-			ConfigType:    "java",
-			ConfigVersion: 1,
+`,
+			want: StaticLauncherConfig{
+				LauncherConfig: LauncherConfig{
+					ConfigType:    "java",
+					ConfigVersion: 1,
+				},
+				Env: map[string]string{
+					"SOME_ENV_VAR":  "/etc/profile",
+					"OTHER_ENV_VAR": "/etc/redhat-release",
+				},
+				Executable: "java",
+				Args:       []string{"arg1", "arg2"},
+				JavaConfig: JavaConfig{
+					MainClass: "mainClass",
+					JavaHome:  "javaHome",
+					Classpath: []string{"classpath1", "classpath2"},
+					JvmOpts:   []string{"jvmOpt1", "jvmOpt2"},
+				},
+			},
 		},
-		MainClass: "mainClass",
-		JavaHome:  "javaHome",
-		Env: map[string]string{
-			"SOME_ENV_VAR":  "/etc/profile",
-			"OTHER_ENV_VAR": "/etc/redhat-release",
+		{
+			name: "executable static config",
+			data: `
+configType: executable
+configVersion: 1
+executable: /usr/bin/postgres
+env:
+  SOME_ENV_VAR: /etc/profile
+  OTHER_ENV_VAR: /etc/redhat-release
+args:
+  - arg1
+  - arg2
+`,
+			want: StaticLauncherConfig{
+				LauncherConfig: LauncherConfig{
+					ConfigType:    "executable",
+					ConfigVersion: 1,
+				},
+				Env: map[string]string{
+					"SOME_ENV_VAR":  "/etc/profile",
+					"OTHER_ENV_VAR": "/etc/redhat-release",
+				},
+				Executable: "/usr/bin/postgres",
+				Args:       []string{"arg1", "arg2"},
+			},
 		},
-		Classpath: []string{"classpath1", "classpath2"},
-		JvmOpts:   []string{"jvmOpt1", "jvmOpt2"},
-		Args:      []string{"arg1", "arg2"},
+	} {
+		got, _ := ParseStaticConfig([]byte(currCase.data))
+		assert.Equal(t, currCase.want, got, "Case %d: %s", i, currCase.name)
 	}
-
-	got := ParseStaticConfig(data)
-	assert.Equal(t, want, got)
 }
 
 func TestParseCustomConfig(t *testing.T) {
@@ -66,7 +104,7 @@ func TestParseCustomConfig(t *testing.T) {
 		want CustomLauncherConfig
 	}{
 		{
-			name: "standard custom config",
+			name: "java custom config",
 			data: `
 configType: java
 configVersion: 1
@@ -90,7 +128,7 @@ jvmOpts:
 			},
 		},
 		{
-			name: "custom config without env",
+			name: "java custom config without env",
 			data: `
 configType: java
 configVersion: 1
@@ -107,7 +145,7 @@ jvmOpts:
 			},
 		},
 		{
-			name: "custom config with env placeholder",
+			name: "java custom config with env placeholder v1",
 			data: `
 configType: java
 configVersion: 1
@@ -128,8 +166,114 @@ jvmOpts:
 				JvmOpts: []string{"jvmOpt1", "jvmOpt2"},
 			},
 		},
+		{
+			name: "executable custom config",
+			data: `
+configType: executable
+configVersion: 1
+env:
+  SOME_ENV_VAR: /etc/profile
+  OTHER_ENV_VAR: /etc/redhat-release
+`,
+			want: CustomLauncherConfig{
+				LauncherConfig: LauncherConfig{
+					ConfigType:    "executable",
+					ConfigVersion: 1,
+				},
+				Env: map[string]string{
+					"SOME_ENV_VAR":  "/etc/profile",
+					"OTHER_ENV_VAR": "/etc/redhat-release",
+				},
+			},
+		},
 	} {
-		got := ParseCustomConfig([]byte(currCase.data))
+		got, _ := ParseCustomConfig([]byte(currCase.data))
 		assert.Equal(t, currCase.want, got, "Case %d: %s", i, currCase.name)
+	}
+}
+
+func TestParseStaticConfigFailures(t *testing.T) {
+	for i, currCase := range []struct {
+		name string
+		msg  string
+		data string
+	}{
+		{
+			name: "bad YAML",
+			msg:  `Failed to deserialize Static Launcher Config, please check the syntax of your configuration file`,
+			data: `
+bad: yaml:
+`,
+		},
+		{
+			name: "invalid config type",
+			msg:  `Can handle configType\=\{.+\} only, found config`,
+			data: `
+configType: config
+configVersion: 1
+executable: postgres
+`,
+		},
+		{
+			name: "invalid config version",
+			msg:  `Can handle configVersion\=\{1\} only, found 2`,
+			data: `
+configType: executable
+configVersion: 2
+executable: postgres
+`,
+		},
+		{
+			name: "invalid executable",
+			msg:  `Can handle executable\=\{.+\} only, found /bin/rm`,
+			data: `
+configType: executable
+configVersion: 1
+executable: /bin/rm
+args:
+  - "-rf"
+  - "/"
+`,
+		},
+		{
+			name: "missing executable",
+			msg:  `Config type \"executable\" requires top-level \"executable:\" value`,
+			data: `
+configType: executable
+configVersion: 1
+`,
+		},
+		{
+			name: "missing java main class and classpath",
+			msg:  `(MainClass|Classpath): zero value`,
+			data: `
+configType: java
+configVersion: 1
+`,
+		},
+		{
+			name: "missing java main class",
+			msg:  `MainClass: zero value`,
+			data: `
+configType: java
+configVersion: 1
+classpath:
+  - thing1
+  - thing2
+`,
+		},
+		{
+			name: "missing java classpath",
+			msg:  `Classpath: zero value`,
+			data: `
+configType: java
+configVersion: 1
+mainClass: hello.world
+`,
+		},
+	} {
+		_, err := ParseStaticConfig([]byte(currCase.data))
+		assert.NotEqual(t, err, nil, "Case %d: %s had no errors", i, currCase.name)
+		assert.Regexp(t, currCase.msg, err.Error(), "Case %d: %s had the wrong error message", i, currCase.name)
 	}
 }
--- a/launchlib/launcher.go
+++ b/launchlib/launcher.go
@@ -30,14 +30,35 @@ const (
 	ExecPathBlackListRegex = `[^\w.\/_\-]`
 )
 
-// Returns true iff the given path is safe to be passed to exec(): must not contain funky characters and be a valid file.
-func verifyPathIsSafeForExec(execPath string) string {
-	unsafe, _ := regexp.MatchString(ExecPathBlackListRegex, execPath)
-	_, statError := os.Stat(execPath)
-	if unsafe || statError != nil {
-		panic("Failed to determine is path is safe to execute: " + execPath)
+func LaunchWithConfig(staticConfigFile, customConfigFile string) error {
+	staticConfig, staticConfigErr := GetStaticConfigFromFile(staticConfigFile)
+	if staticConfigErr != nil {
+		return staticConfigErr
 	}
-	return execPath
+
+	customConfig, customConfigErr := GetCustomConfigFromFile(customConfigFile)
+	if customConfigErr != nil {
+		return customConfigErr
+	}
+
+	launchErr := Launch(&staticConfig, &customConfig)
+	if launchErr != nil {
+		return launchErr
+	}
+	return nil
+}
+
+// Returns true iff the given path is safe to be passed to exec(): must not contain funky characters and be a valid file.
+func verifyPathIsSafeForExec(execPath string) (string, error) {
+	if unsafe, err := regexp.MatchString(ExecPathBlackListRegex, execPath); err != nil {
+		return "", err
+	} else if unsafe {
+		return "", fmt.Errorf("Unsafe execution path: %q ", execPath)
+	} else if _, statErr := os.Stat(execPath); statErr != nil {
+		return "", statErr
+	}
+
+	return execPath, nil
 }
 
 type processExecutor interface {
@@ -48,16 +69,16 @@ type syscallProcessExecutor struct{}
 
 // Returns explicitJavaHome if it is not the empty string, or the value of the JAVA_HOME environment variable otherwise.
 // Panics if neither of them is set.
-func getJavaHome(explicitJavaHome string) string {
+func getJavaHome(explicitJavaHome string) (string, error) {
 	if explicitJavaHome != "" {
-		return explicitJavaHome
+		return explicitJavaHome, nil
 	}
 
 	javaHome := os.Getenv("JAVA_HOME")
 	if len(javaHome) == 0 {
-		panic("JAVA_HOME environment variable not set")
+		return "", fmt.Errorf("JAVA_HOME environment variable not set")
 	}
-	return javaHome
+	return javaHome, nil
 }
 
 func getWorkingDir() string {
@@ -81,43 +102,64 @@ func joinClasspathEntries(classpathEntries []string) string {
 	return strings.Join(classpathEntries, ":")
 }
 
-func Launch(staticConfig *StaticLauncherConfig, customConfig *CustomLauncherConfig) {
+func Launch(staticConfig *StaticLauncherConfig, customConfig *CustomLauncherConfig) error {
 	fmt.Printf("Launching with static configuration %v and custom configuration %v\n", *staticConfig, *customConfig)
 
 	workingDir := getWorkingDir()
 	fmt.Println("Working directory:", workingDir)
 
-	javaHome := getJavaHome(staticConfig.JavaHome)
-	fmt.Println("Using JAVA_HOME:", javaHome)
-	javaCommand := verifyPathIsSafeForExec(path.Join(javaHome, "/bin/java"))
-
-	classpath := joinClasspathEntries(absolutizeClasspathEntries(workingDir, staticConfig.Classpath))
-	fmt.Println("Classpath:", classpath)
-
 	var args []string
-	args = append(args, javaCommand) // 0th argument is the command itself
-	args = append(args, staticConfig.JvmOpts...)
-	args = append(args, customConfig.JvmOpts...)
-	args = append(args, "-classpath", classpath)
-	args = append(args, staticConfig.MainClass)
+	var executable string
+	var executableErr error
+
+	if staticConfig.ConfigType == "java" {
+		javaHome, javaHomeErr := getJavaHome(staticConfig.JavaConfig.JavaHome)
+		if javaHomeErr != nil {
+			return javaHomeErr
+		}
+		fmt.Println("Using JAVA_HOME:", javaHome)
+
+		classpath := joinClasspathEntries(absolutizeClasspathEntries(workingDir, staticConfig.JavaConfig.Classpath))
+		fmt.Println("Classpath:", classpath)
+
+		executable, executableErr = verifyPathIsSafeForExec(path.Join(javaHome, "/bin/java"))
+		if executableErr != nil {
+			return executableErr
+		}
+		args = append(args, executable) // 0th argument is the command itself
+		args = append(args, staticConfig.JavaConfig.JvmOpts...)
+		args = append(args, customConfig.JvmOpts...)
+		args = append(args, "-classpath", classpath)
+		args = append(args, staticConfig.JavaConfig.MainClass)
+	} else if staticConfig.ConfigType == "executable" {
+		executable, executableErr = verifyPathIsSafeForExec(staticConfig.Executable)
+		if executableErr != nil {
+			return executableErr
+		}
+		args = append(args, executable) // 0th argument is the command itself
+	} else {
+		return fmt.Errorf("You can't launch type %v, this should have errored in config validation", staticConfig.ConfigType)
+	}
+
 	args = append(args, staticConfig.Args...)
-	fmt.Printf("Argument list to Java binary: %v\n\n", args)
+	fmt.Printf("Argument list to executable binary: %v\n\n", args)
 
 	env := replaceEnvironmentVariables(merge(staticConfig.Env, customConfig.Env))
 
-	execWithChecks(javaCommand, args, env, &syscallProcessExecutor{})
+	execWithChecks(executable, args, env, &syscallProcessExecutor{})
+	return nil
 }
 
-func execWithChecks(javaExecutable string, args []string, customEnv map[string]string, p processExecutor) {
+func execWithChecks(executable string, args []string, customEnv map[string]string, p processExecutor) {
 	env := os.Environ()
 	for key, value := range customEnv {
 		env = append(env, fmt.Sprintf("%s=%s", key, value))
 	}
 
-	execErr := p.Exec(javaExecutable, args, env)
+	execErr := p.Exec(executable, args, env)
 	if execErr != nil {
 		if os.IsNotExist(execErr) {
-			fmt.Println("Java Executable not found at:", javaExecutable)
+			fmt.Println("Executable not found at:", executable)
 		}
 		panic(execErr)
 	}
--- a/launchlib/launcher_test.go
+++ b/launchlib/launcher_test.go
@@ -34,10 +34,12 @@ func TestGetJavaHome(t *testing.T) {
 	originalJavaHome := os.Getenv("JAVA_HOME")
 	require.NoError(t, os.Setenv("JAVA_HOME", "foo"))
 
-	javaHome := getJavaHome("")
+	javaHome, javaHomeErr := getJavaHome("")
 	assert.Equal(t, "foo", javaHome, "JAVA_HOME incorrect")
-	javaHome = getJavaHome("explicit javahome")
+	assert.NoError(t, javaHomeErr, "getJavaHome correctly returns nil")
+	javaHome, javaHomeErr = getJavaHome("explicit javahome")
 	assert.Equal(t, "explicit javahome", javaHome, "JAVA_HOME incorrect")
+	assert.NoError(t, javaHomeErr, "getJavaHome correctly returns nil")
 
 	require.NoError(t, os.Setenv("JAVA_HOME", originalJavaHome))
 }
--- a/vendor/github.com/pkg/errors/LICENSE
+++ b/vendor/github.com/pkg/errors/LICENSE
@@ -1,0 +1,23 @@
+Copyright (c) 2015, Dave Cheney <dave@cheney.net>
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
+* Redistributions in binary form must reproduce the above copyright notice,
+  this list of conditions and the following disclaimer in the documentation
+  and/or other materials provided with the distribution.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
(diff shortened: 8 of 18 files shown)


main.go
```diff
@@ -19,7 +19,7 @@ import (
 	"io/ioutil"
 	"os"
 
-	"github.com/palantir/go-java-launcher/launchlib"
+	"github.palantir.build/tkozlowski/go-launcher/launchlib"
```
uschi2000: oops

uschi2000 (commented):
it's not clear that we really want to expand the scope of this library. can you explain the rationale?

bmoylan:
Discussion on rationale should probably happen on an issue before creating a PR

main.go
tomkozlowski: :facepalm:

uschi2000:
I don't think we can take this change; the purpose of this library securely parametrised execution of (Java) programs. Since this change enables bash execution, it enables arbitrary code execution via parameter injection.

---

Review status: 0 of 9 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-:-KYTvjvyVAOdJUPy10iN:bsuqfif)*

tomkozlowski:
@uschi2000 so, it don't think this actually does allow for arbitrary param injection, the env variables aren't actually evaluated if referenced in arguments to the program. Also, I think I have mislabeled this as a `shell` execution env, as we never actually hit a shell during execution, instead we make a syscall directly https://github.com/palantir/go-java-launcher/pull/19/files#diff-0dd5d40edbd03a1d824891bcffcce657R128

uschi2000:
sorry, was mislead by the PR title earlier. re-visiting, I think we can take this. only concern is that we should refactor the code eventually to account for the java/executable configType branching more elegantly. concerns, @markelliot

markelliot:
It seems like one could configure bash to be the base command and then templated arguments would execute arbitrary code. That's not functionality I'd be comfortable with. 

Could you back up a bit and explain the goals of what you're trying to accomplish and how this implementation is intended to support that?

uschi2000:
Yes, one could do that, but the same attack vector exists for Java: I can
write an evil program, launch it with go-java-launcher and have it evaluate
my command line arguments. Or am I missing something?

markelliot:
Our deployment system protects which artifacts can be shipped, so the hole would require going through our publishing and deploy system. This would allow someone with access to just the UI to move forward.

Either way, I'd like to understand the rationale before continuing the thread.

tomkozlowski:
@markelliot i'm trying to generalize the go launcher so that I can run other executables (postgres, influx, grafana) with the same protections that java gets. the next step I'm planning to take is writing in initialization support so that the launcher wrapper handles initialization (e.g. postgres's initdb execution) in a more secure manner rather than using the bash scripts I have currently

markelliot:
Do those executables typically take lots of arguments?

tomkozlowski:
grafana takes 7, influx takes 3, postgres takes 5

but the bigger drive behind my efforts is to get the ability to do secure initialization and I see this PR as a necessary stepping stone to that goal.

tomkozlowski:
@uschi2000 I'm done with the rebase and rename, do you want to do another pass?

uschi2000:
Should ensure Mark's concerns are addressed, then I'll take a second pass

tomkozlowski:
@markelliot would explicitly blacklisting shell executables (and I suppose `sshd` as well) be an acceptable mitigation?

Or, alternatively we could stick to explicitly whitelisted executables (postgres/influx/grafana/etc.), but I could see that generating more churn on the repo

integration_test/test_resources/launcher-custom.yml
```diff
@@ -1,4 +1,5 @@
 configType: java
-configVersion: 1
-jvmOpts:
-  - '-Xmx1g'
+configVersion: 2
```
markelliot: so we're going to need to make sure we continue support configVersion 1

markelliot:
I don't think it's effective to be able to blacklist commands. As an alternative, we could whitelist specific executables, but that's probably headache that also doesn't work. The main thing I'd object to about this is that the static config ends up specifying the executable to run, of course it is the static config and we trust it in a number of other ways.

tomkozlowski:
My read is that this project so far works on the whitelisting principle with only ``java`` whitelisted.  

I would think that having the executable only definable in the static config, this wouldn't expand the existing attack surface per @uschi2000 's [comment](https://github.com/palantir/go-java-launcher/pull/19#issuecomment-265778900) that the existing code is still susceptible to someone writing a malicious Java program. In a sense, it seems like we're trusting developers to not bundle static configs with malicious (or potentially malicious) executables (in the existing case, java programs).

integration_test/test_resources/launcher-custom.yml
tomkozlowski: yeah, that should be fairly easy to do. I think the exact implementation will depend on how we decide to handle generic executables vs. explicitly whitelisted executables

uschi2000:
I don't quite follow the java-vs-executable security concern: in the case
of the Java launcher, the launcher-static file contains the classpath,
i.e., effectively the code to be executed. This could in turn decide to run
Postgres or Bash. So how is the executable mechanism proposed in this PR
different?

tomkozlowski:
@uschi2000 it's not different, that's my point. Or rather, my point is that this doesn't introduce additional vulnerabilities (such as RCE for someone with web config access) by restricting executable and arg definitions to the static conf

bmoylan:
As long as the executable is not configurable at runtime and the arguments are not subject to shell expansion I think this is reasonable. Java is special because we need the classpath to be immutable (the executable alone doesn't limit what code gets run)

I can currently package anything I want to run in an `init.sh` and we consider it safe because it is immutable. So setting the executable to `sshd` in `init.sh` vs. `launcher-static.yml` require the same access. Setting the executable to a shell (`sh`, `bash`, `zsh`, etc) would be bad because then the configurable arguments give you remote shell access. This _could_ be enforced by a blacklist but I'm sure we won't be able to prevent all stupidity of this kind.

markelliot:
Ok, I withdraw my opposition on the scope of the change, but want to flag that:

1.       We don't want to have to migrate existing launcher-static.yml files, back-versions need to be supported in some effective way

2.       I think we should consider a whitelist of program names

tomkozlowski:
@markelliot to your points:

1. Yep, I amended the PR to maintain backcompat with v1
2. Do you think this consideration should block merging this PR or do you think it should be tracked separately?
(and adding my own question) 3. One of the changes that is currently staged is changing the name of the binary from `go-java-launcher` to `go-launcher` which would also imply that we would want to change the name of the repo (given that it now supports executables other than just go), where do you stand on this?

markelliot:
1.       Cool

2.       I think we should decide whitelist now or never; renaming artifact + repo isn't an issue, nobody has customized those things (launcher-static + launcher-custom have been customized, though, which is why I care)

tomkozlowski:
@uschi2000 / @markelliot alright, i think this is ready for an actual CR, can one of y'all check it out on reviewable?

bmoylan:
High-level CR comments: 
- we should be returning errors instead of panicking everywhere. The panics make the code really hard to follow.
- It should happen in a separate commit, but we should move to `github.com/palantir/pkg/cli` now that it's open-source. Do you mind opening a ticket for that?

---

Reviewed 1 of 10 files at r1, 9 of 13 files at r2.
Review status: 9 of 10 files reviewed at latest revision, 12 unresolved discussions.

---

*[README.md, line 6 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoAMvj7qMzAawLUw10:-KYoAMvj7qMzAawLUw11:btdvy4p) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/README.md#L6)):*
> ```Markdown
> # go-launcher
> 
> A simple Go program for launching  programs from a fixed configuration. This program replaces Gradle-generated Bash
> ```

nit: extra space after `launching`

---

*[README.md, line 16 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoBAz0LPRDb55m0t7u:-KYoBAz0LPRDb55m0t7v:b-wx01mq) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/README.md#L16)):*
> ```Markdown
> # The type of configuration, must be the string "java" or "executable"
> configType: java
> # The version of the configuration format, must be the integer 1 or 2
> ```

It's not clear here the difference between v1 and v2 or when to use which

---

*[README.md, line 31 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoAg-tie8R2Yqv_5aa:-KYoAg-tie8R2Yqv_5ab:b7njhcz) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/README.md#L31)):*
> ```Markdown
> jvmOpts:
>   - '-Xmx1g'
> # The full path to the executable file, ignored if configType is "java", limited to whitelisted values (java, postgres, influxd, grafana-server)
> ```

This needs to support relative paths just like the classpath, right? You don't know where it will be deployed on disk.

---

*[launchlib/config.go, line 42 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoUV6sNIAwF567iAPQ:-KYoUV6tdZl8kBFOKaVk:b42qhgg) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/config.go#L42)):*
> ```Go
> 
> func (config *LauncherConfig) validateLauncherConfig(){
> 	_, typeOk := allowedLauncherConfigs.ConfigTypes[config.ConfigType]
> ```

these checks can be inlined

```
if _, ok := allowedLauncherConfigs.ConfigTypes[config.ConfigType]; !ok {
   // error
}
```

---

*[launchlib/config.go, line 43 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoUTHHMbeUdbqiLAEQ:-KYoUTHHMbeUdbqiLAER:b-nvuwzo) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/config.go#L43)):*
> ```Go
> func (config *LauncherConfig) validateLauncherConfig(){
> 	_, typeOk := allowedLauncherConfigs.ConfigTypes[config.ConfigType]
> 	_, versionOk := allowedLauncherConfigs.ConfigVersions[config.ConfigVersion]
> ```

move this down to where you check it

---

*[launchlib/config.go, line 48 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoV7ciB8P9v5NUPiYO:-KYoV7ciB8P9v5NUPiYP:b-4sbc3p) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/config.go#L48)):*
> ```Go
> 		for k := range allowedLauncherConfigs.ConfigTypes { allowedConfigTypes = append(allowedConfigTypes, k) }
> 
> 		panic(fmt.Sprintf("Can handle configType=%v only, found %v", allowedConfigTypes, config.ConfigType))
> ```

prefer `%s` when you know the value is a string (like `ConfigType`)

---

*[launchlib/config.go, line 60 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoVFmm2wqf1T67wW78:-KYoVFmm2wqf1T67wW79:b9nzumr) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/config.go#L60)):*
> ```Go
> type JavaConfig struct {
> 	JavaHome      string            `yaml:"javaHome"`
> 	MainClass     string            `yaml:"mainClass"`
> ```

formatting looks off here, make sure to run `./godelw format` and commit the results

---

*[launchlib/config.go, line 88 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoXLV6hC8EY4Bd1ScQ:-KYoXLV6hC8EY4Bd1ScR:bz054el) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/config.go#L88)):*
> ```Go
> }
> 
> func validateJavaConfig(javaConfig JavaConfig){
> ```

you can use `validate:"nonzero"` on the struct tags instead of needing this

---

*[launchlib/config.go, line 95 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoWeJGu9mnRoN9k2Ne:-KYoWeJGu9mnRoN9k2Nf:bqc2z0z) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/config.go#L95)):*
> ```Go
> 
> func validateExecutableConfig(executable string){
> 	if len(executable) <= 0 {
> ```

`len` will never be `<0`(also `exectutable == ""` is more idiomatic)

---

*[launchlib/config.go, line 114 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoXdBTFVcfTMeL0PAQ:-KYoXdBU2J5yfwsM3H8B:blw91ld) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/config.go#L114)):*
> ```Go
> 	if config.ConfigType == "java" {
> 		validateJavaConfig(config.JavaConfig)
> 		config.Executable = "java"
> ```

this assumes `java` is on the PATH, don't we want to use `$JAVA_HOME/bin/java`?

---

*[launchlib/launcher.go, line 124 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoYuQU-uyKPdptwrAU:-KYoYuQU-uyKPdptwrAV:b-ga7292) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/launcher.go#L124)):*
> ```Go
> 		args = append(args, "-classpath", classpath)
> 		args = append(args, staticConfig.JavaConfig.MainClass)
> 	} else if staticConfig.ConfigType == "executable" {
> ```

nit: thoughts on `exec` instead of `executable`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-:-KYoBzmuxd95pTzTEzaD:b-hpjmxq)*

tomkozlowski:
Review status: 9 of 10 files reviewed at latest revision, 12 unresolved discussions.

---

*[main.go, line 22 at r1](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYR6Wr-DaFkVSIOfY9i-r1-22:-KYoY1xd3oiwibOlNZJj:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/f9e5cd3fe8b1128d4e49faa71dcad6eab0660928/main.go#L22)):*
<details><summary><i>Previously, tomkozlowski (Tom K) wrote…</i></summary><blockquote>

:facepalm:
</blockquote></details>

Done.

---

*[README.md, line 6 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoAMvj7qMzAawLUw10:-KYo_3bOX2NOC_TJuxDY:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/README.md#L6)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

nit: extra space after `launching`
</blockquote></details>

Done.

---

*[README.md, line 16 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoBAz0LPRDb55m0t7u:-KYo_65g_BstxBue04GJ:bq43r0e) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/README.md#L16)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

It's not clear here the difference between v1 and v2 or when to use which
</blockquote></details>

v2 adds support for specifying a value for `executable`, but it's a superset of the v1 functionality, so I could easily just eliminate the concept of v2

---

*[README.md, line 31 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoAg-tie8R2Yqv_5aa:-KYo_OyUStof_ShYDO0i:b-qxyqxa) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/README.md#L31)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

This needs to support relative paths just like the classpath, right? You don't know where it will be deployed on disk.
</blockquote></details>

Yeah, per the bottom of the README, I envision this generally being `{{CWD}}/service/bin/whatever`, that could be spelled out more explicitly

---

*[launchlib/config.go, line 42 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoUV6sNIAwF567iAPQ:-KYo_iUVqUkRvgnXa0t5:b-7jc4ze) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/config.go#L42)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

these checks can be inlined

```
if _, ok := allowedLauncherConfigs.ConfigTypes[config.ConfigType]; !ok {
   // error
}
```
</blockquote></details>

solid

---

*[launchlib/config.go, line 48 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoV7ciB8P9v5NUPiYO:-KYo_rL9jkYaRNlm-0tM:bb7m07t) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/config.go#L48)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

prefer `%s` when you know the value is a string (like `ConfigType`)
</blockquote></details>

okay

---

*[launchlib/config.go, line 60 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoVFmm2wqf1T67wW78:-KYo_tK0CAwhhvKuwNhy:bq0ncok) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/config.go#L60)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

formatting looks off here, make sure to run `./godelw format` and commit the results
</blockquote></details>

sounds good

---

*[launchlib/config.go, line 88 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoXLV6hC8EY4Bd1ScQ:-KYo_wrZg-fIw9oUMVkD:b-kdtps3) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/config.go#L88)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

you can use `validate:"nonzero"` on the struct tags instead of needing this
</blockquote></details>

didn't know that

---

*[launchlib/config.go, line 95 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoWeJGu9mnRoN9k2Ne:-KYoa0LjuriqJoY2CFXR:b-agu0sc) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/config.go#L95)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

`len` will never be `<0`(also `exectutable == ""` is more idiomatic)
</blockquote></details>

alright, i can go with the more idiomatic approach. I decided to go with the logical opposite of `> 0` even though length is by definition never negative

---

*[launchlib/config.go, line 114 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoXdBTFVcfTMeL0PAQ:-KYoaEUzNeIdmqMhS24_:b6c4s6s) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/config.go#L114)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

this assumes `java` is on the PATH, don't we want to use `$JAVA_HOME/bin/java`?
</blockquote></details>

no, that never actually calls `java`, it's there to validate against the approved executable list

---

*[launchlib/launcher.go, line 124 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoYuQU-uyKPdptwrAU:-KYoa_DuXLyoISygJxK7:b-sot8yd) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/launcher.go#L124)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

nit: thoughts on `exec` instead of `executable`?
</blockquote></details>

i defaulted to the more precise name, but I'm not that dogmatic on that

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/go-java-launcher/19)*

tomkozlowski:
Review status: 9 of 10 files reviewed at latest revision, 12 unresolved discussions.

---

*[launchlib/config.go, line 42 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoUV6sNIAwF567iAPQ:-KYoeMO0ossv5es0aTd_:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/config.go#L42)):*
<details><summary><i>Previously, tomkozlowski (Tom K) wrote…</i></summary><blockquote>

solid
</blockquote></details>

Done.

---

*[launchlib/config.go, line 43 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoUTHHMbeUdbqiLAEQ:-KYodmYDF4xCWO59ILTf:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/config.go#L43)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

move this down to where you check it
</blockquote></details>

Done.

---

*[launchlib/config.go, line 48 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoV7ciB8P9v5NUPiYO:-KYoeNAjbUHsy8mpNCwP:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/config.go#L48)):*
<details><summary><i>Previously, tomkozlowski (Tom K) wrote…</i></summary><blockquote>

okay
</blockquote></details>

Done.

---

*[launchlib/config.go, line 60 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoVFmm2wqf1T67wW78:-KYohzA6kOLBbLKp-M2g:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/config.go#L60)):*
<details><summary><i>Previously, tomkozlowski (Tom K) wrote…</i></summary><blockquote>

sounds good
</blockquote></details>

Done.

---

*[launchlib/config.go, line 88 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoXLV6hC8EY4Bd1ScQ:-KYohxwimATIkz0ZUad6:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/config.go#L88)):*
<details><summary><i>Previously, tomkozlowski (Tom K) wrote…</i></summary><blockquote>

didn't know that
</blockquote></details>

Done.

---

*[launchlib/config.go, line 95 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoWeJGu9mnRoN9k2Ne:-KYohxJw7fVpZoqwehJa:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/config.go#L95)):*
<details><summary><i>Previously, tomkozlowski (Tom K) wrote…</i></summary><blockquote>

alright, i can go with the more idiomatic approach. I decided to go with the logical opposite of `> 0` even though length is by definition never negative
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/go-java-launcher/19)*

bmoylan:
Review status: 6 of 10 files reviewed at latest revision, 4 unresolved discussions.

---

*[README.md, line 31 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoAg-tie8R2Yqv_5aa:-KYoi_QJR6j8cezx25zk:bmnvrp3) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/README.md#L31)):*
<details><summary><i>Previously, tomkozlowski (Tom K) wrote…</i></summary><blockquote>

Yeah, per the bottom of the README, I envision this generally being `{{CWD}}/service/bin/whatever`, that could be spelled out more explicitly
</blockquote></details>

Do we actually support `{{CWD}}` as a template expansion? Are there other thing supported like this?

---

*[launchlib/config.go, line 114 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoXdBTFVcfTMeL0PAQ:-KYojkRQGDY57OzO7zhS:b32w0x) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/launchlib/config.go#L114)):*
<details><summary><i>Previously, tomkozlowski (Tom K) wrote…</i></summary><blockquote>

no, that never actually calls `java`, it's there to validate against the approved executable list
</blockquote></details>

OK

---

*[launchlib/config.go, line 69 at r3](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoj7KlME7GbFU_Dkio:-KYoj7Ko8U136p-yEzTi:bgs11p6) ([raw file](https://github.com/palantir/go-java-launcher/blob/51c143fe01bdb44c708120eb14e503518faba697/launchlib/config.go#L69)):*
> ```Go
> type StaticLauncherConfig struct {
> 	LauncherConfig `yaml:",inline"`
> 	JavaConfig     `yaml:",inline",validate:"nonzero"`
> ```

I think you'll need it on each of the individual fields of `JavaConfig`. When you get rid of all the panics the negative tests will help you out here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/go-java-launcher/19)*

tomkozlowski:
Review status: 6 of 10 files reviewed at latest revision, 4 unresolved discussions.

---

*[README.md, line 31 at r2](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoAg-tie8R2Yqv_5aa:-KYonaV265t-i243LDXZ:b-x1dltu) ([raw file](https://github.com/palantir/go-java-launcher/blob/87ed5d0ad8191c9e751c142cc7fe5a89366e31b0/README.md#L31)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

Do we actually support `{{CWD}}` as a template expansion? Are there other thing supported like this?
</blockquote></details>

Based on the bottom of the readme, just `{{CWD}}` for now

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/go-java-launcher/19)*

tomkozlowski:
Review status: 4 of 9 files reviewed at latest revision, 4 unresolved discussions.

---

*[launchlib/config.go, line 69 at r3](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYoj7KlME7GbFU_Dkio:-KYsum0Ff0gXd_LWU2yi:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/51c143fe01bdb44c708120eb14e503518faba697/launchlib/config.go#L69)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

I think you'll need it on each of the individual fields of `JavaConfig`. When you get rid of all the panics the negative tests will help you out here.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/go-java-launcher/19)*

uschi2000:
Review status: 4 of 9 files reviewed at latest revision, 6 unresolved discussions.

---

*[main.go, line 29 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYui5CoQaJ_Ow-pdNlP:-KYui5Cp4W4NNgHct9Ad:bu0q398) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/main.go#L29)):*
> ```Go
> 	switch numArgs := len(os.Args); {
> 	case numArgs > 3:
> 		panic("Usage: go-java-launcher [<path to StaticLauncherConfig> [<path to CustomLauncherConfig>]]")
> ```

inconsistent

---

*[integration_test/integration_test.go, line 47 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYuiaDvujNxvm1L923R:-KYuiaDvujNxvm1L923S:b-n43kat) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/integration_test/integration_test.go#L47)):*
> ```Go
> 	// part of expected output from launcher
> 	assert.Regexp(t, `Failed to read custom config file, assuming no custom config: foo`, output)
> 	assert.Regexp(t, `Argument list to Java binary: \[.+/bin/java -Xmx4M -classpath .+/github.com/palantir/go-launcher/integration_test/test_resources Main arg1\]`, output)
> ```

don't quite understand how this can work since we haven't renamed the repository yet?

---

*[launchlib/config.go, line 70 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYul16oXO_NmKmhTycx:-KYul16oXO_NmKmhTycy:b-uq4uh0) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L70)):*
> ```Go
> }
> 
> func validateJavaConfig(config JavaConfig) error {
> ```

don't quite understand the file structure; why not group the validateXyz functions?

---

*[launchlib/config.go, line 99 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYul_LyeBnuhrXp1GCx:-KYul_LyeBnuhrXp1GCy:bxbz106) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L99)):*
> ```Go
> 		return fmt.Errorf("Config type \"executable\" requires top-level \"executable:\" value")
> 	}
> 	if _, executableOk := allowedLauncherConfigs.Executables[path.Base(executable)]; !executableOk {
> ```

the same structural code for turning an array into a collection appears multiple times in this file

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/go-java-launcher/19)*

uschi2000:
Review status: 4 of 9 files reviewed at latest revision, 7 unresolved discussions.

---

*[README.md, line 10 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYumIanCLGi0PVD3hl1:-KYumIanCLGi0PVD3hl2:bwx50iz) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/README.md#L10)):*
> ```Markdown
> /)'`.
> 
> The launcher accepts as configuration two YAML files as follows:
> ```

the semantics of "type = java" vs "type = executable" are not clear from reading this. would propose to separate these into two different listings, one explaining how to launch java things, one explaining how to launch executable things.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/go-java-launcher/19)*

bmoylan:
Reviewed 3 of 4 files at r3, 1 of 5 files at r4, 3 of 3 files at r5.
Review status: all files reviewed at latest revision, 20 unresolved discussions.

---

*[launchlib/config.go, line 111 at r4](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYtkoKT4UlX0pdfx-vK:-KYtkoKT4UlX0pdfx-vL:b-6qtwtz) ([raw file](https://github.com/palantir/go-java-launcher/blob/722601a27b9aef4803dbdaaa22049b068250eaab/launchlib/config.go#L111)):*
> ```Go
> 	var config StaticLauncherConfig
> 	if err := yaml.Unmarshal(yamlString, &config); err != nil {
> 		return StaticLauncherConfig{}, fmt.Errorf("Failed to deserialize Static Launcher Config, please check the syntax of your configuration file\n")
> ```

You're dropping the original error from Unmarshal that has all the context. You want something like `errors.Wrap(err, "Failed to deserialize Static Launcher Config, please check the syntax of your configuration file")` 

(`errors` is `github.com/pkg/errors` if it's not vendored yet)

---

*[launchlib/config.go, line 115 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw5cH5jDGBkG3jHhr9:-KYw5cH5jDGBkG3jHhrA:b-8u25hk) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L115)):*
> ```Go
> 		return StaticLauncherConfig{}, fmt.Errorf("Failed to deserialize Static Launcher Config, please check the syntax of your configuration file\n")
> 	}
> 	if launcherConfigOk := config.LauncherConfig.validateLauncherConfig(); launcherConfigOk != nil {
> ```

call the error `err`; things with `ok` are assumed to be booleans

---

*[launchlib/config.go, line 120 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw5iF545lsvpRmkF7l:-KYw5iF545lsvpRmkF7m:b-1rn2ql) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L120)):*
> ```Go
> 
> 	if config.ConfigType == "java" {
> 		if javaConfigOk := validateJavaConfig(config.JavaConfig); javaConfigOk != nil {
> ```

call it `err`

---

*[launchlib/config.go, line 126 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw5lLgVj5hXjQQts8W:-KYw5lLgVj5hXjQQts8X:b-1rn2ql) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L126)):*
> ```Go
> 	}
> 
> 	if executableConfigOk := validateExecutableConfig(config.Executable); executableConfigOk != nil {
> ```

call it `err`

---

*[launchlib/config.go, line 135 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw59_FQaTU7CzBp7C0:-KYw59_FQaTU7CzBp7C1:b-77j0i8) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L135)):*
> ```Go
> 	staticData, err := ioutil.ReadFile(staticConfigFile)
> 	if err != nil {
> 		return StaticLauncherConfig{}, fmt.Errorf("Failed to read static config file: " + staticConfigFile)
> ```

`errors.Wrap` to not lose `err`

---

*[launchlib/config.go, line 137 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw5q5lXcKPjIp9GzdW:-KYw5q5lXcKPjIp9GzdX:bqkp3ef) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L137)):*
> ```Go
> 		return StaticLauncherConfig{}, fmt.Errorf("Failed to read static config file: " + staticConfigFile)
> 	}
> 	staticConfig, staticConfigOk := ParseStaticConfig(staticData)
> ```

call the error `err`

---

*[launchlib/config.go, line 147 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw5txvmM7BOIY-lDXy:-KYw5txvmM7BOIY-lDXz:b-dqf34i) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L147)):*
> ```Go
> 	var config CustomLauncherConfig
> 	if err := yaml.Unmarshal(yamlString, &config); err != nil {
> 		return CustomLauncherConfig{}, fmt.Errorf("Failed to deserialize Custom Launcher Config, please check the syntax of your configuration file\n")
> ```

`errors.Wrap`

---

*[launchlib/config.go, line 149 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw5wTEZ6mB9m1jWnAc:-KYw5wTFXteIDYW0j2gn:bqkp3ef) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L149)):*
> ```Go
> 		return CustomLauncherConfig{}, fmt.Errorf("Failed to deserialize Custom Launcher Config, please check the syntax of your configuration file\n")
> 	}
> 	if launcherConfigOk := config.LauncherConfig.validateLauncherConfig(); launcherConfigOk != nil {
> ```

call the error `err`

---

*[launchlib/config.go, line 160 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw61ThJXnglPbUcVZj:-KYw61ThJXnglPbUcVZk:bqkp3ef) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L160)):*
> ```Go
> 		return CustomLauncherConfig{}, nil
> 	} else {
> 		customConfig, customConfigOk := ParseCustomConfig(customData)
> ```

call the error `err`

---

*[launchlib/config_test.go, line 195 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw6qEDdqoU7Dprdvzf:-KYw6qEDdqoU7Dprdvzg:basqv1o) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config_test.go#L195)):*
> ```Go
> }
> 
> func TestParseStaticConfigFailures(t *testing.T) {
> ```

bonus points for checking the error message is correct

---

*[launchlib/launcher.go, line 35 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw6NZ2PgJ6JOgtX9EA:-KYw6NZ2PgJ6JOgtX9EB:b-1mjud4) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/launcher.go#L35)):*
> ```Go
> func exit1WithMessage(message string) {
> 	fmt.Fprintln(os.Stderr, message)
> 	os.Exit((1))
> ```

In general you don't want to hide `os.Exit` places. I'd make `LaunchWithConfig` return an error and check it in main.go

---

*[launchlib/launcher.go, line 39 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw6auyOyfpaLP7NhVv:-KYw6auyOyfpaLP7NhVw:blvvjmx) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/launcher.go#L39)):*
> ```Go
> 
> func LaunchWithConfig(staticConfigFile, customConfigFile string) {
> 	staticConfig, staticConfigOk := GetStaticConfigFromFile(staticConfigFile)
> ```

call the error `err` and return it

---

*[launchlib/launcher.go, line 44 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw6cvu7d9dZHNAhZCq:-KYw6cvu7d9dZHNAhZCr:blvvjmx) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/launcher.go#L44)):*
> ```Go
> 	}
> 
> 	customConfig, customConfigOk := GetCustomConfigFromFile(customConfigFile)
> ```

call the error `err` and return it

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/go-java-launcher/19)*

tomkozlowski:
Review status: all files reviewed at latest revision, 20 unresolved discussions.

---

*[main.go, line 29 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYui5CoQaJ_Ow-pdNlP:-KYxr_10VL_rYuoOlwol:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/main.go#L29)):*
<details><summary><i>Previously, uschi2000 (Robert Fink) wrote…</i></summary><blockquote>

inconsistent
</blockquote></details>

Done.

---

*[README.md, line 10 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYumIanCLGi0PVD3hl1:-KYxs6pRVb1RjtoW8pnJ:bieztpq) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/README.md#L10)):*
<details><summary><i>Previously, uschi2000 (Robert Fink) wrote…</i></summary><blockquote>

the semantics of "type = java" vs "type = executable" are not clear from reading this. would propose to separate these into two different listings, one explaining how to launch java things, one explaining how to launch executable things.
</blockquote></details>

Good point, I'll fix that

---

*[integration_test/integration_test.go, line 47 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYuiaDvujNxvm1L923R:-KYxwW2wZ1GIVvfgC0vE:bg5wvh3) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/integration_test/integration_test.go#L47)):*
<details><summary><i>Previously, uschi2000 (Robert Fink) wrote…</i></summary><blockquote>

don't quite understand how this can work since we haven't renamed the repository yet?
</blockquote></details>

I defer to @bmoylan, but I think we have a chicken/egg problem with renaming reference paths and renaming the repo, so I chose to do the rename of reference paths in the last commit of my PR

---

*[launchlib/config.go, line 111 at r4](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYtkoKT4UlX0pdfx-vK:-KYy5X0zYZ-jRuVm6u4M:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/722601a27b9aef4803dbdaaa22049b068250eaab/launchlib/config.go#L111)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

You're dropping the original error from Unmarshal that has all the context. You want something like `errors.Wrap(err, "Failed to deserialize Static Launcher Config, please check the syntax of your configuration file")` 

(`errors` is `github.com/pkg/errors` if it's not vendored yet)
</blockquote></details>

Done.

---

*[launchlib/config.go, line 70 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYul16oXO_NmKmhTycx:-KYxwlKnxGArkhjZEV30:b-8j7xlb) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L70)):*
<details><summary><i>Previously, uschi2000 (Robert Fink) wrote…</i></summary><blockquote>

don't quite understand the file structure; why not group the validateXyz functions?
</blockquote></details>

I was under the impression that golang idioms were to place struct validation functions under the struct definition rather than grouping validate funcs together, but I defer to @bmoylan as the expert.

---

*[launchlib/config.go, line 99 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYul_LyeBnuhrXp1GCx:-KYy56d4o_Vb5av8vUtB:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L99)):*
<details><summary><i>Previously, uschi2000 (Robert Fink) wrote…</i></summary><blockquote>

the same structural code for turning an array into a collection appears multiple times in this file
</blockquote></details>

Done.

---

*[launchlib/config.go, line 115 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw5cH5jDGBkG3jHhr9:-KYy5yMU9Zdf7DSA7Cd6:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L115)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

call the error `err`; things with `ok` are assumed to be booleans
</blockquote></details>

Done.

---

*[launchlib/config.go, line 120 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw5iF545lsvpRmkF7l:-KYy5yjY-aT5Km5GSdZn:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L120)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

call it `err`
</blockquote></details>

Done.

---

*[launchlib/config.go, line 126 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw5lLgVj5hXjQQts8W:-KYy5zTyC1GVn7ObVydI:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L126)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

call it `err`
</blockquote></details>

Done.

---

*[launchlib/config.go, line 135 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw59_FQaTU7CzBp7C0:-KYy68cty9_vWRqIN-Wf:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L135)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

`errors.Wrap` to not lose `err`
</blockquote></details>

Done.

---

*[launchlib/config.go, line 137 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw5q5lXcKPjIp9GzdW:-KYy6KqY45UHS3WODm9C:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L137)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

call the error `err`
</blockquote></details>

Done.

---

*[launchlib/config.go, line 147 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw5txvmM7BOIY-lDXy:-KYy6QvsnET-Zuam6IEf:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L147)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

`errors.Wrap`
</blockquote></details>

Done.

---

*[launchlib/config.go, line 149 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw5wTEZ6mB9m1jWnAc:-KYy6RiEtgGqvfrcxfFz:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L149)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

call the error `err`
</blockquote></details>

Done.

---

*[launchlib/config.go, line 160 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw61ThJXnglPbUcVZj:-KYy6SOUQyGSMH_odqQu:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L160)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

call the error `err`
</blockquote></details>

Done.

---

*[launchlib/config_test.go, line 195 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw6qEDdqoU7Dprdvzf:-KYy7ypHHFBHH46KNldl:bgj6187) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config_test.go#L195)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

bonus points for checking the error message is correct
</blockquote></details>

ugh, i'm not running for Pope here, damn you for making me write good code

---

*[launchlib/launcher.go, line 35 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw6NZ2PgJ6JOgtX9EA:-KYyBz_azk4C9zUpzRAE:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/launcher.go#L35)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

In general you don't want to hide `os.Exit` places. I'd make `LaunchWithConfig` return an error and check it in main.go
</blockquote></details>

Done.

---

*[launchlib/launcher.go, line 39 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw6auyOyfpaLP7NhVv:-KYyC-pSl3RcisfAY0RW:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/launcher.go#L39)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

call the error `err` and return it
</blockquote></details>

Done.

---

*[launchlib/launcher.go, line 44 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw6cvu7d9dZHNAhZCq:-KYyC0HwTVbsQLDiIQlm:b-896fix) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/launcher.go#L44)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

call the error `err` and return it
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/go-java-launcher/19)*

bmoylan:
Make sure to run `./godelw verify` yourself - because this is building a fork CircleCI isn't running the tests for you

---

Reviewed 5 of 5 files at r6.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[launchlib/config.go, line 70 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYul16oXO_NmKmhTycx:-KYyeomT-cSJOErVd-1T:beroomp) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L70)):*
<details><summary><i>Previously, tomkozlowski (Tom K) wrote…</i></summary><blockquote>

I was under the impression that golang idioms were to place struct validation functions under the struct definition rather than grouping validate funcs together, but I defer to @bmoylan as the expert.
</blockquote></details>

I don't have strong opinions here, but the order you normally see in a go file is 

1. Exported structs
2. unexported structs
3. Exported funcs
4. unexported funcs

Instead of having this validation function at all, though, you can make the JavaConfig struct's unmarshaller do it for you with the `validate` tag:

```
type JavaConfig struct {
	JavaHome  string   `yaml:"javaHome"`
	MainClass string   `yaml:"mainClass" validate:"nonzero"`
	JvmOpts   []string `yaml:"jvmOpts"`
	Classpath []string `yaml:"classpath" validate:"nonzero"`
}
```

---

*[launchlib/config.go, line 115 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw5cH5jDGBkG3jHhr9:-KYyWMVuRCajtfNs4UDP:bobyuy9) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L115)):*
<details><summary><i>Previously, tomkozlowski (Tom K) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

the errors don't each need their own name - it's shorter and more idiomatic to just call them all `err` because they're used in a very small scope (this applies to the everywhere in the PR)

---

*[launchlib/config.go, line 83 at r6](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYyebQDCbSpPsZFwGyC:-KYyebQDCbSpPsZFwGyD:b-6wawuy) ([raw file](https://github.com/palantir/go-java-launcher/blob/981e8101de831bbde25dda53d4b998415f6f1154/launchlib/config.go#L83)):*
> ```Go
> func validateJavaConfig(config JavaConfig) error {
> 	if config.MainClass == "" {
> 		return fmt.Errorf("Config type \"java\" requires top-level \"mainClass:\" value")
> ```

use `fmt.Error` (no `f`) if it's not a format string (there are a few of these and I _think_ golint will complain when you `godel verify`)

---

*[launchlib/config.go, line 120 at r6](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYyWCcGm6fdXQVqydeO:-KYyWCcH9XHALktXpLkQ:bupti0r) ([raw file](https://github.com/palantir/go-java-launcher/blob/981e8101de831bbde25dda53d4b998415f6f1154/launchlib/config.go#L120)):*
> ```Go
> 	var config StaticLauncherConfig
> 	if err := yaml.Unmarshal(yamlString, &config); err != nil {
> 		return StaticLauncherConfig{}, errors.Wrap(err, "Failed to deserialize Static Launcher Config, please check the syntax of your configuration file\n")
> ```

don't need the newline on the end

---

*[launchlib/launcher.go, line 58 at r6](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYyae9tcdunctdzyOZt:-KYyae9tcdunctdzyOZu:b-neihk4) ([raw file](https://github.com/palantir/go-java-launcher/blob/981e8101de831bbde25dda53d4b998415f6f1154/launchlib/launcher.go#L58)):*
> ```Go
> 	_, statError := os.Stat(execPath)
> 	if unsafe {
> 		return "", errors.Wrap(unsafeErr, "Failed to determine is path is safe to execute: "+execPath)
> ```

Careful here - there's a difference between `unsafe` being true (meaning the regex matched) and `unsafeErr` being non-nil (meaning there was an error running the regex itself). You probably want

```
unsafe, err := regexp.MatchString(ExecPathBlackListRegex, execPath)
if err != nil {
  return "", err
}
if unsafe {
  return "", fmt.Errorf("unsafe path %q", execPath)
}
if _, err := os.Stat(execPath); err != nil {
  return "", err
}
```

---

*[launchlib/launcher_test.go, line 39 at r6](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYyeM_3VkuHAlYPuJq6:-KYyeM_3VkuHAlYPuJq7:blij3yb) ([raw file](https://github.com/palantir/go-java-launcher/blob/981e8101de831bbde25dda53d4b998415f6f1154/launchlib/launcher_test.go#L39)):*
> ```Go
> 	javaHome, javaHomeErr := getJavaHome("")
> 	assert.Equal(t, "foo", javaHome, "JAVA_HOME incorrect")
> 	assert.Nil(t, javaHomeErr, "getJavaHome correctly returns nil")
> ```

`assert.NoError` gives a slightly nicer message

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-:-KYy_Q8C5T-ofxPGy2GI:b-l1p2nf)*

tomkozlowski:
done

---

Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[launchlib/config.go, line 70 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYul16oXO_NmKmhTycx:-KYyij3f5gDm6J9ps-jv:b-hr49qv) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L70)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

I don't have strong opinions here, but the order you normally see in a go file is 

1. Exported structs
2. unexported structs
3. Exported funcs
4. unexported funcs

Instead of having this validation function at all, though, you can make the JavaConfig struct's unmarshaller do it for you with the `validate` tag:

```
type JavaConfig struct {
	JavaHome  string   `yaml:"javaHome"`
	MainClass string   `yaml:"mainClass" validate:"nonzero"`
	JvmOpts   []string `yaml:"jvmOpts"`
	Classpath []string `yaml:"classpath" validate:"nonzero"`
}
```
</blockquote></details>

Aight, reordering based on what Brad said and using the validate thing

---

*[launchlib/config.go, line 115 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYw5cH5jDGBkG3jHhr9:-KYytCtKPDibeeTbhbUq:b-7n88fv) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/launchlib/config.go#L115)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

the errors don't each need their own name - it's shorter and more idiomatic to just call them all `err` because they're used in a very small scope (this applies to the everywhere in the PR)
</blockquote></details>

aight, aligning with this based on your more detailed explanation on slack

---

*[launchlib/config.go, line 83 at r6](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYyebQDCbSpPsZFwGyC:-KYynBlhzQLPJ3YM949p:b-bhxvz8) ([raw file](https://github.com/palantir/go-java-launcher/blob/981e8101de831bbde25dda53d4b998415f6f1154/launchlib/config.go#L83)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

use `fmt.Error` (no `f`) if it's not a format string (there are a few of these and I _think_ golint will complain when you `godel verify`)
</blockquote></details>

`fmt.Error` doesn't exist - https://golang.org/pkg/fmt/#Errorf, so i'll use `errors.New`

---

*[launchlib/config.go, line 120 at r6](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYyWCcGm6fdXQVqydeO:-KYynQVwZt8RWlYqufOm:b-kn5lj8) ([raw file](https://github.com/palantir/go-java-launcher/blob/981e8101de831bbde25dda53d4b998415f6f1154/launchlib/config.go#L120)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

don't need the newline on the end
</blockquote></details>

I think that was a legacy of the old usage of panics, removing

---

*[launchlib/launcher.go, line 58 at r6](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYyae9tcdunctdzyOZt:-KYyufDDyibe-ZxGebX0:b-otbuus) ([raw file](https://github.com/palantir/go-java-launcher/blob/981e8101de831bbde25dda53d4b998415f6f1154/launchlib/launcher.go#L58)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

Careful here - there's a difference between `unsafe` being true (meaning the regex matched) and `unsafeErr` being non-nil (meaning there was an error running the regex itself). You probably want

```
unsafe, err := regexp.MatchString(ExecPathBlackListRegex, execPath)
if err != nil {
  return "", err
}
if unsafe {
  return "", fmt.Errorf("unsafe path %q", execPath)
}
if _, err := os.Stat(execPath); err != nil {
  return "", err
}
```


</blockquote></details>

I went a slightly different route, but accomplishing the same nuance

---

*[launchlib/launcher_test.go, line 39 at r6](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYyeM_3VkuHAlYPuJq6:-KYyulRrvNBD6n_xcg7r:boyexq7) ([raw file](https://github.com/palantir/go-java-launcher/blob/981e8101de831bbde25dda53d4b998415f6f1154/launchlib/launcher_test.go#L39)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

`assert.NoError` gives a slightly nicer message
</blockquote></details>

fine....

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-:-KYywWCNi0xFsOPHAowP:bb7f2mf)*

bmoylan:
:lgtm: thanks for putting up with my nitpicks! can also confirm that tests/verify are passing locally for me

---

Reviewed 18 of 20 files at r7.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[integration_test/integration_test.go, line 47 at r5](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYuiaDvujNxvm1L923R:-KYz6BDKTI__tES7RMJ3:b-v4gs0h) ([raw file](https://github.com/palantir/go-java-launcher/blob/b08c8f9d923989455f0653b470eeaf7656fe7e07/integration_test/integration_test.go#L47)):*
<details><summary><i>Previously, tomkozlowski (Tom K) wrote…</i></summary><blockquote>

I defer to @bmoylan, but I think we have a chicken/egg problem with renaming reference paths and renaming the repo, so I chose to do the rename of reference paths in the last commit of my PR
</blockquote></details>

Decided on Slack to do the rename in a different PR because it needs to be coordinated with GitHub settings changes (and possibly other things like bintray?)

I'll leave it up to you two to coordinate that, but it should happen in the same _release_ as this PR.

---

*[vendor/vendor.json, line 13 at r7](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYz9BspAapLMO0nkCuc:-KYz9BspAapLMO0nkCud:b-ajqt6t) ([raw file](https://github.com/palantir/go-java-launcher/blob/c5e203e5bc30cbc9dde5c44dc5dacd24a3e6ad4b/vendor/vendor.json#L13)):*
> ```JSON
> 		},
> 		{
> 			"path": "github.com/palantir/go-java-launcher/launchlib",
> ```

this may have gotten in here during the rename but you don't actually need it in vendor.json

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-:-KYz9dFs8iSw05UeOdPe:bzfykx2)*

tomkozlowski:
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[vendor/vendor.json, line 13 at r7](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-KYz9BspAapLMO0nkCuc:-KYzk1E_CEQN-8hv-JPb:bd5ke7e) ([raw file](https://github.com/palantir/go-java-launcher/blob/c5e203e5bc30cbc9dde5c44dc5dacd24a3e6ad4b/vendor/vendor.json#L13)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

this may have gotten in here during the rename but you don't actually need it in vendor.json
</blockquote></details>

is it fine to leave in there?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/go-java-launcher/19)*

tomkozlowski:
@uschi2000 can you take another look on reviewable and approve/ack your open questions and give it your +1 unless you have other thoughts?

uschi2000:
:lgtm:

---

Reviewed 18 of 20 files at r7.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/go-java-launcher/19#-:-KYzy2oHRiLmyc4H7t9b:bnfp4nl)*

tpetracca:
@tomkozlowski does this change when coupled with `init.sh start` from `sls-packaging` record the proper pid in the pid file? I assume `service/bin/postgres` is just a wrapper that launches yet another process?

tomkozlowski:
I'm not familiar with the `sls-packaging` change you're referring to, but this should work with postgres since postgres spins off a number of child processes, but the main PID should be captured correctly for the master process

jared2501:
Looking to use this for launching envoy and it appears the whitelist prohibits this. I'm not sure if the whitelisting is actually necessary here considering that launcher-static.yml is statically defined in the artifact. @markelliot - would you be open to a change removing this whitelist?

jared2501:
Or at least blacklisting bash/sh/zsh/etc.

markelliot:
I’d be fine extending the whitelist to include envoy, but not to switch to a blacklist.

jared2501:
Why?

jared2501:
Note that the executable would be defined in "launcher-static.yml" which is never templated from our configuration editing UI.

jared2501:
The blacklisting/whitelisting functionality is purely so that people don't shoot themselves in the foot and launch bash scripts.

markelliot:
We chose a whitelist because exhaustively listing risky processes seems like a losing battle, and we had (and continue to have) a knowable and relatively small set of processes we’re willing to bless.

markelliot:
It’s more than a foot shooting guard — one cannot simply launch _any_ process. Why is this minor friction too high for you?

PR #26, opened 2017-02-15
launchlib returns a exec.Cmd struct rather than launching itself

uschi2000:
This change is [Reviewable](https://reviewable.io/reviews/palantir/go-java-launcher/26)


uschi2000:
@nmiyake let me know if anything's broken here. will move on with life until until then.

nmiyake:
Yup took a look and this change looks fine

PR #30, opened 2017-03-13
Rfink/init start

uschi2000:
This change is [Reviewable](https://reviewable.io/reviews/palantir/go-java-launcher/30)

### DIFF
--- a/init/cli/start.go
+++ b/init/cli/start.go
@@ -33,18 +33,21 @@ Runs the command defined by the given static and custom configurations and store
 the given pid file.
 `,
 		Flags: []flag.Flag{
-			flag.StringParam{
+			flag.StringFlag{
 				Name:  launcherStaticFileParameter,
-				Usage: ""},
-			flag.StringParam{
+				Value: "service/bin/launcher-static.yml",
+				Usage: "The location of the LauncherStatic file configuration the started command"},
+			flag.StringFlag{
 				Name:  launcherCustomFileParameter,
-				Usage: ""},
-			flag.StringParam{
+				Usage: "The location of the LauncherCustom file configuration the started command"},
+			flag.StringFlag{
 				Name:  pidfileParameter,
-				Usage: ""},
-			flag.StringParam{
+				Value: "var/run/service.pid",
+				Usage: "The location of the file storing the process ID of the started command"},
+			flag.StringFlag{
 				Name:  outFileParameter,
-				Usage: ""},
+				Value: "var/log/startup.log",
+				Usage: "The location of the file to which STDOUT and STDERR of the started command are redirected"},
 		},
 		Action: doStart,
 	}
--- a/init/cli/start.go
+++ b/init/cli/start.go
@@ -39,6 +39,7 @@ the given pid file.
 				Usage: "The location of the LauncherStatic file configuration the started command"},
 			flag.StringFlag{
 				Name:  launcherCustomFileParameter,
+				Value: "var/conf/launcher-custom.yml",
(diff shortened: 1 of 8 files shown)


markelliot:
lgtm, only thought is that we might want to make it so that it automatically discovers a default-located launcher-custom if one exists so that one can just run `go-java-launcher start`.

uschi2000:
meh, that was a copy&paste typo, thanks.

init/cli/start.go
```diff
@@ -0,0 +1,85 @@
...
+		Name: "start",
+		Usage: `
+Runs the command defined by the given static and custom configurations and stores the PID of the resulting process in
+the given pid file.
+`,
+		Flags: []flag.Flag{
+			flag.StringFlag{
+				Name:  launcherStaticFileParameter,
+				Value: "service/bin/launcher-static.yml",
+				Usage: "The location of the LauncherStatic file configuration the started command"},
```
nmiyake, line 39: Typo: `file configuration the started command` should be `of the started command`

init/cli/start.go
```diff
@@ -0,0 +1,85 @@
...
+			flag.StringFlag{
+				Name:  launcherCustomFileParameter,
+				Value: "var/conf/launcher-custom.yml",
+				Usage: "The location of the LauncherCustom file configuration the started command"},
```
nmiyake, line 43: Same typo as above

init/cli/start.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	originalStdout := os.Stdout
+	os.Stdout = stdoutfile // log command assembly output to file instead of stdout
+	cmd, err := launchlib.CompileCmdFromConfigFiles(launcherStaticFile, launcherCustomFile)
+	if err != nil {
+		msg := fmt.Sprintln("Failed to assemble Command object from static and custom configuration files", err)
+		return respondError(msg, err, 1)
+	}
+	os.Stdout = originalStdout
```
nmiyake, line 76: Is your goal to always restore the original `stdout` before function exit? If so, this should be done as: 

```
defer func() {
   os.Stdout = originalStdout
}()
```

immediately after the original assignment to ensure that it is restored. Otherwise, non-standard return paths will not restore `stdout` (for example, right now if `CompileCmdFromConfigFiles` returns a non-nil error, `stdout` is not restored before function exit)

init/cli/status.go
```diff
@@ -35,9 +31,10 @@ Determines the status of the process whose PID is contained in the given pid-fil
 process is running, 1 if the pid-file exists but the process is not running, and 3 if the pid-file
 does not exist`,
 		Flags: []flag.Flag{
-			flag.StringParam{
+			flag.StringFlag{
 				Name:  pidfileParameter,
-				Usage: "The path to a file containing the PID of for which the status is to be determined"},
+				Usage: "The path to a file containing the PID of for which the status is to be determined",
```
nmiyake, line 36: Typo: `of for which the...`

nmiyake (commented):
Some post-merge comments. Mostly surface-level, but may take a closer look later.

init/cli/start.go
uschi2000: done

init/cli/start.go
uschi2000: done

init/cli/start.go
uschi2000: thanks. done

init/cli/status.go
uschi2000: done

PR #33, opened 2017-04-23
Add support for creating directories

### LINKED ISSUE #32: Make Java tmp directory by default

uschi2000:
Fixes #32

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/go-java-launcher/33)

### DIFF
--- a/integration_test/integration_test.go
+++ b/integration_test/integration_test.go
@@ -15,6 +15,7 @@
 package integration_test
 
 import (
+	"os"
 	"os/exec"
 	"testing"
 
@@ -47,6 +48,21 @@ func TestMainMethodWithoutCustomConfig(t *testing.T) {
 	assert.Regexp(t, `Argument list to executable binary: \[.+/bin/java -Xmx4M -classpath .+/github.com/palantir/go-java-launcher/integration_test/test_resources Main arg1\]`, output)
 	// expected output of Java program
 	assert.Regexp(t, `\nmain method\n`, string(output))
+}
+
+func TestCreatesDirs(t *testing.T) {
+	output, err := runMainWithArgs(t, "test_resources/launcher-static-with-dirs.yml", "foo")
+	require.NoError(t, err, "failed: %s", output)
+
+	dir, err := os.Stat("foo")
+	assert.NoError(t, err)
+	assert.True(t, dir.IsDir())
+	require.NoError(t, os.RemoveAll("foo"))
+
+	dir, err = os.Stat("bar/baz")
+	assert.NoError(t, err)
+	assert.True(t, dir.IsDir())
+	require.NoError(t, os.RemoveAll("bar"))
 }
 
 func runMainWithArgs(t *testing.T, staticConfigFile, customConfigFile string) (string, error) {
--- a/launcher/main/main.go
+++ b/launcher/main/main.go
@@ -41,12 +41,32 @@ func main() {
 		customConfigFile = os.Args[2]
 	}
 
-	cmd, err := launchlib.CompileCmdFromConfigFiles(staticConfigFile, customConfigFile)
+	// Read configuration
+	staticConfig, err := launchlib.GetStaticConfigFromFile(staticConfigFile)
+	if err != nil {
+		fmt.Println("Failed to read static config file", err)
+		panic(err)
+	}
+	customConfig, err := launchlib.GetCustomConfigFromFile(customConfigFile)
+	if err != nil {
+		fmt.Println("Failed to read custom config file", err)
+		panic(err)
+	}
+
+	// Create configured directories
+	err = launchlib.MkDirs(staticConfig.Dirs)
+	if err != nil {
+		fmt.Println("Failed to create directories", err)
+	}
+
+	// Compile command
+	cmd, err := launchlib.CompileCmdFromConfig(&staticConfig, &customConfig)
 	if err != nil {
 		fmt.Println("Failed to assemble executable metadata", cmd, err)
 		panic(err)
 	}
 
+	// Execute command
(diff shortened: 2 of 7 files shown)


integration_test/integration_test.go
```diff
@@ -49,6 +50,21 @@ func TestMainMethodWithoutCustomConfig(t *testing.T) {
...
 }
 
+func TestCreatesDirs(t *testing.T) {
+	output, err := runMainWithArgs(t, "test_resources/launcher-static-with-dirs.yml", "foo")
+	require.NoError(t, err, "failed: %s", output)
+
+	dir, err := os.Stat("foo")
+	assert.NoError(t, err)
+	assert.True(t, dir.IsDir())
+	os.RemoveAll("foo")
```
nmiyake: Store `err` output of this and `require.NoError`

integration_test/integration_test.go
```diff
@@ -49,6 +50,21 @@ func TestMainMethodWithoutCustomConfig(t *testing.T) {
...
+
+	dir, err = os.Stat("bar/baz")
+	assert.NoError(t, err)
+	assert.True(t, dir.IsDir())
+	os.RemoveAll("bar")
```
nmiyake: Same

launcher/main/main.go
```diff
@@ -41,12 +41,29 @@ func main() {
...
+		panic(err)
+	}
+	customConfig, err := launchlib.GetCustomConfigFromFile(customConfigFile)
+	if err != nil {
+		fmt.Println("Failed to read custom config file", err)
+		panic(err)
+	}
+
+	// Create configured directories
+	launchlib.MkDirs(staticConfig.Dirs)
```
nmiyake: Check error output

launchlib/launcher.go
```diff
@@ -91,6 +91,22 @@ func CompileCmdFromConfig(staticConfig *StaticLauncherConfig, customConfig *Cust
 	return createCmd(executable, args, env)
 }
 
+func MkDirs(dirs []string) error {
+	isDirMatcher := regexp.MustCompile(`^[A-Za-z0-9]+(/[A-Za-z0-9]+)*$`).MatchString
```
nmiyake, line 95: Probably fine in this case, but in long-lived applications it's typically better to store the result of `regexp.MustCompile` in a package-private variable so that it's parsed/compiled once

launchlib/launcher.go
```diff
@@ -91,6 +91,22 @@ func CompileCmdFromConfig(staticConfig *StaticLauncherConfig, customConfig *Cust
...
+	for _, dir := range dirs {
+		if !isDirMatcher(dir) {
+			return fmt.Errorf("Cannot create directory with non [A-Za-z0-9] characters: %s", dir)
+		}
+
+		fmt.Printf("Creating directory: %s\n", dir)
+		err := os.MkdirAll(dir, 0700)
```
nmiyake: Minor matter of preference, but if a function returns only an error that you check just once, you can inline the call and check (and this is often done in Go):

```go
if err := os.MkdirAll(dir, 0700); err != nil {
  return err
}
```

launchlib/launcher_test.go
```diff
@@ -112,3 +112,22 @@ func TestKeysAreNotExpanded(t *testing.T) {
...
+	err := MkDirs([]string{"abc/def1"})
+	assert.NoError(t, err)
+
+	err = MkDirs([]string{"abc"})
+	assert.NoError(t, err)
+
+	require.NoError(t, os.RemoveAll("abc"))
+
+	err = MkDirs([]string{"^&*"})
+	assert.Error(t, err)
```
nmiyake: If you're checking for a known error, typically better to assert something about the error message here (ensures that this test only passes if it fails for the reason you expect rather than for any reason)

launcher/main/main.go
```diff
@@ -41,12 +41,32 @@ func main() {
...
+	customConfig, err := launchlib.GetCustomConfigFromFile(customConfigFile)
+	if err != nil {
+		fmt.Println("Failed to read custom config file", err)
+		panic(err)
+	}
+
+	// Create configured directories
+	err = launchlib.MkDirs(staticConfig.Dirs)
+	if err != nil {
+		fmt.Println("Failed to create directories", err)
```
nmiyake: Right now, operation continues after directory creation fails -- is this intentional, or should you panic/return?

launcher/main/main.go
```diff
@@ -41,12 +41,32 @@ func main() {
...
+		panic(err)
+	}
+	customConfig, err := launchlib.GetCustomConfigFromFile(customConfigFile)
+	if err != nil {
+		fmt.Println("Failed to read custom config file", err)
+		panic(err)
+	}
+
+	// Create configured directories
+	err = launchlib.MkDirs(staticConfig.Dirs)
```
nmiyake: Can combine call and check per comment further below, but purely a matter of preference.

nmiyake (commented):
Some minor feedback, but core logic of the change looks good.

launchlib/launcher.go
uschi2000: done

launchlib/launcher.go
uschi2000: this is single-use, so i prefer the readability in this case. thanks!

launcher/main/main.go
uschi2000: should panic, thanks.

launcher/main/main.go
uschi2000: done

launchlib/launcher_test.go
uschi2000: done

PR #36, opened 2017-05-30
Use CircleCI 2.0 for builds

nmiyake:


nmiyake:
Ah right the integration tests require Java to exist as well, which would require creating a Docker image with Go + Java. Not too hard to do, but will tackle some other time.

PR #37, opened 2017-08-26
Allow additional classpath args in custom configs.

billbindi:
This allows users to add adtional classpath args in the custom config
file in the same vein as adding additional jvmOpts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ dirs:
 ```yaml
 # StaticLauncherConfig - executable version
 # REQUIRED - The type of configuration, must be the string "executable"
-configType: java
+configType: executable
 # REQUIRED - The version of the configuration format, must be the integer 1
 configVersion: 1
 # OPTIONAL - Environment Variables to be set in the environment (Note: cannot be referenced on args list)
@@ -70,6 +70,10 @@ env:
 # Additional JVM options to be passed to the java command, will override defaults in static config. Ignored if configType is "executable"
 jvmOpts:
   - '-Xmx2g'
+# Additional classpath entries to be appended to the existing classpath list in the static config.
+# Append only. Unlike the static config, this must be an absolute path to the resource.
+classpath
+  - /path/to/additional/resource
 ```
 
 The launcher is invoked as:
--- a/integration_test/init_integration_test.go
+++ b/integration_test/init_integration_test.go
@@ -36,8 +36,8 @@ func TestInitStatus(t *testing.T) {
 	// No valid pidfile
 	stdout, stderr, exitCode := runInit("status", "--pidFile", "bogus-file")
 	assert.Empty(t, stdout)
-	assert.Equal(t, stderr, "Failed to determine whether process is running for pid-file: bogus-file. Exit code: 3. "+
-		"Underlying error: open bogus-file: no such file or directory")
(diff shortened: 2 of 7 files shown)


palantirtech:
Thanks for your interest in [palantir/go-java-launcher](https://github.com/palantir/go-java-launcher), @billbindi! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

billbindi:
@nmiyake @uschi2000 I don't know how to add reviewers, but I would like this if possible.

uschi2000:
This library is explicitly designed around the idea of not allowing runtime changes to the classpath. This change is at odds with that idea, I think?

billbindi:
What exactly do you mean by "runtime" changes? This will all happen at launch time, the same way the jvm options are customizable. If this is still undesired, then do you have a recommendation for swapping between multiple jars for different runs. More specifically, if I want to have a "plugin" (which is just a jar file) for my application be implemented by two completely disjoint jars, how do I set the runner to pick one over the other? Am I just supposed to modify the static config file and not the custom one?

bmoylan:
SLS is built on the idea of immutable distributions -- trying to mutate the classpath at runtime (aka launch time, aka install time) is an antipattern that isn't going to be supported well by any of the tooling. It's also a scary vector for untrusted code execution if the deployment platform is compromised.

billbindi:
Just to be clear, mutating the jvm args is supported (including overwriting existing ones), but adding to the classpath is not? I definitely understand that one can be more intrusive than the other, but is the distinction really that strong? I would be interested to hear of a case where the platform is compromised and not having this proposed change would mean that you are safe from an attack. Even redeploying the service (if you knew that it was compromised) won't be sufficient if there are cronjobs and other nefarious things left around by the attacker - same as if they could modify the custom file, no?

billbindi:
Closing for now since this is not desired. I would be curious to know if there is a recommendation for the above scenario (i.e. a legitimate way of switching out jars for an SLS service), but for now we will continue to do it manually.

billbindi:
Sorry, I would like to revisit this. It looks like in the jvmOpts you are not barred from using an arbitrary java agent. So the custom launcher is already a vulnerability if someone has access to it, no? Perhaps allowing the changing of the classpath directly is more direct, but I don't think you are really guarding against custom code execution if you allow the specification of arbitrary jvm args.

I understand the idea of the immutable distribution, and I am all for that. In my head, this feature would be used not by encouraging people to force in jars into the laid out application, but rather to add a jar from a known, absolute path (e.g. `/home/user/jars/my-custom-plugin.jar`), and the config file is what actually adds in this code. I am not sure what tooling you had in mind, but if this is still too strong of an option for that, I won't push the issue too hard, but this seems like it could be quite helpful for some services that want the ability to swap in plugins that were not around at build/distribution time.

billbindi:
Just wanted to ping this and see if you have given any more thought to merging this? If it is really undesired I will back off, but I am not sure I fully understand the drawbacks here.

uschi2000:
I want to maintain the separation between compile-time and runtime, and in particular the requirement to fix the class path at compile time. You're right in saying that -D arguments can be used to do whatever, but I would consider that a bug, not a feature.

Sent from Email+ secured by MobileIron

-------- Original Message --------

billbindi:
Closing for good

PR #40, opened 2017-11-17
Allow non-home tildes on exec path

### LINKED ISSUE #39: Allow for tildes on exec path if they aren't the start of the path
We use tildes in install paths (`/foo/bar/baz/group~value~id`). This doesn't jive well with the current regex: https://github.com/palantir/go-java-launcher/blob/develop/launchlib/launcher.go#L110 when the provided `JAVA_HOME` is a subdirectory of that install path.

Request is to allow tildes if they aren't at the beginning of the path (AKA won't resolve to the home directory).

munkyshi:
Fixes https://github.com/palantir/go-java-launcher/issues/39

### DIFF
--- a/launchlib/launcher.go
+++ b/launchlib/launcher.go
@@ -27,7 +27,7 @@ const (
 	TemplateDelimsOpen  = "{{"
 	TemplateDelimsClose = "}}"
 	// ExecPathBlackListRegex matches characters disallowed in paths we allow to be passed to exec()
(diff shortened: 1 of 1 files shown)


uschi2000:
What's the rationale for this change?

jlong8:
Check out the description in https://github.com/palantir/go-java-launcher/issues/39; we use tildes in install paths, and this blacklist currently makes those paths invalid.

PR #41, opened 2018-04-17
add yourkit to custom config and set java args

jnelson15:
\+ @iamdanfox @tstearns


jnelson15:
Maybe worth checking if the file exists and just logging a warning if so rather than failing?

Edit: Actually maybe not since the default is off so we probably don't want a silent fail if someone is explicitly trying to enable this.

iamdanfox:
Chatted over slack - we're gonna shelve this for now as users can unblocked themselves by adding the correct incantation to their existing launcher-custom:

```
-agentpath:./service/lib/linux-x86-64/libyjpagent.so=builtinprobes=none
```

jnelson15:
Talked to Dan, going to close now since (a) we can just set this in the java arg and don't have to hard code paths here and (b) we may change how the infra is done here anyway

PR #45, opened 2018-06-08
go-init

gregakinman:
@cbrockington 

enables https://github.com/palantir/sls-packaging/issues/164

deadcode is failing because I have a function that's only used in test code, which is allowed in deadcode if you specify `-test`, but it looks like we don't do this. @nmiyake is the way to add arguments to a task to follow the example here? https://github.com/palantir/godel-okgo-asset-errcheck/blob/master/errcheck/config/internal/v0/config.go#L23

Also, it seems like `integration_test/doc.go` exists solely so that there exists at least one non-test source file in the `integration_test` package so that `go install $(./godelw packages)` can succeed, which seems silly - if this is the case, probably better for `./godelw packages` to accept a flag to not produce packages that don't include non-test code (opened https://github.com/palantir/godel/issues/401)

Can I get write access to this repo so I can debug with SSH? Some of the tests rely on `pgrep` and work just fine on my machine but on Circle the calls to `pgrep` exit 1 for some reason...


iamdanfox:
@gregakinman that issue has been dormant since Jan 2017... what's the motivation behind switching from bash -> go now?

(Write access granted, so hopefully CircleCI SSH will work for you now)

init/cli/cli.go
```diff
@@ -18,29 +18,11 @@ import (
...
-
 func App() *cli.App {
 	app := cli.NewApp()
 	app.Name = "go-init"
-	app.Usage = "A simple init.sh - style service launcher CLI"
-	app.ErrorHandler = handleError
+	app.Usage = "A simple init.sh-style service launcher CLI."
 
-	app.Subcommands = []cli.Command{statusCommand(), startCommand()}
+	app.Subcommands = []cli.Command{startCommand(), statusCommand(), stopCommand()}
```
iamdanfox, line 26: I think sls-packaging supports a `console` command: https://github.com/palantir/sls-packaging/pull/29

Also, wasn't there functionality to discover git-style subcommand scripts?

cbrockington:
It seems like a lot of the changes in this PR start by just removing a bunch of defaulted command line options.  Is there any particular reason for that?  Their inclusion enables some much simpler end-to-end testing (since you can point your test run to files without having to ensure a bunch of exact locations first), and I'm not sure what we gain by removing them all.

Regardless, if you wish to remove all of the options, could we split this into a "no-op PR" which removes all of the configurability, then a separate PR with all of the additions in functionality (though it's currently unclear to me what these are? which is part of the reason the separation would be nice).

cbrockington:
@iamdanfox the reason for the revitalisation of the bash -> go move is to more easily support more complex process creation logic such as the addition of "sidecar processes" in some legacy environments.

init/cli/cli.go
gregakinman: I explicitly only want to support `start`, `stop`, and `status`, since those are the only arguments that are in the spec. `search?utf8=✓&q="init.sh+console"&type=` on the end of the internal repo base URL shows that `console` isn't used by the tooling, so it seems fine to remove. We can have a conversation about individual cases for the subcommand scripts if those come up, but I'd want a really good reason to include stuff that's outside of the spec.

gregakinman:
@cbrockington I removed the command-line options since, as you noted, it seems like they're there just for testing, and I think in general that adding configurability in production to facilitate testing is a bad thing (it's an artifact of your testing leaking through to production - if there's no clear use case for a property's configurability in production, don't support it if you can help it). Additionally, I can't envision a use case for the flags in production, enforcing a standard for where pidfiles, etc. live is a good thing (and testing that this works when inside that standard setup), and in the current bash world this is the way it is already and it's been fine this whole time. It didn't add much complexity to the tests to set up the SLS distribution structure, so I removed the flags.

It's going to be a fair bit of work to tease apart the changes. I can do it if you really need, but I think it would make it easier for review purposes to look at this PR as completely adding the `init/` directory since it changed so much of what was there. In short, though, the prior state was there was start/status functionality, a few integration tests, and unit tests for start. I added stop and finished adding tests for everything. I also refactored the error reporting to rely on `errors.Wrap` and the `withExitCode` functionality that comes OOTB with the cli package rather than use custom error types with a custom handler.

Also would appreciate a golang-focused review from somebody who knows it well - first time writing Go, so would be good to know if I'm doing anything blatantly wrong/non-idiomatically.

@iamdanfox hmm, I still see this in Circle:
screen shot 2018-06-11 at 15 21 12

nmiyake:
Sorry, I've been caught in some high-pri work from Friday until now so haven't been able to take a look at this yet. That work has mostly wrapped up so I will review this later today or tomorrow for feedback on content and Golang stuff.

init/cli/start.go
```diff
@@ -15,73 +15,48 @@
...
 }
 
-func doStart(ctx cli.Context) error {
-	launcherStaticFile := ctx.String(launcherStaticFileParameter)
-	launcherCustomFile := ctx.String(launcherCustomFileParameter)
-	pidfile := ctx.String(pidfileParameter)
-	stdoutFileName := ctx.String(outFileParameter)
+func start(_ cli.Context) error {
+	_, _, err := lib.GetProcessStatus()
+	if err == nil {
```
nmiyake: nit: in a case like this, inline assignment and check:

```go
if _, _, err := lib.GetProcessStatus(); err == nil {
   return nil
}
```

init/cli/start.go
```diff
@@ -15,73 +15,48 @@
...
+		return nil
+	}
 
-	stdoutFile, err := os.Create(stdoutFileName)
+	outputFile, err := os.Create(lib.OutputFile)
 	if err != nil {
-		msg := fmt.Sprintln("Failed to create startup log file", err)
-		return respondError(msg, err, 1)
+		err = errors.Wrap(err, "failed to create startup log file")
+		return cli.WithExitCode(1, err)
```
nmiyake: Now that the code is restructured, you can just inline both these calls (it makes the code more readable/closer to idiomatic Go):

```go
return cli.WithExitCode(1, errors.Wrap(err, ...))
```

init/cli/start.go
```diff
@@ -15,73 +15,48 @@
...
-			flag.StringFlag{
-				Name:  outFileParameter,
-				Value: "var/log/startup.log",
-				Usage: "The location of the file to which STDOUT and STDERR of the started command are redirected"},
-		},
-		Action: doStart,
+Launches the process defined by the static and custom configurations at service/bin/launcher-static.yml and
+var/conf/launcher-custom.yml. Writes its PID to var/run/service.pid and redirects its output to var/log/startup.log. If
+successful, exits 0, otherwise exits 1 and writes an error message to stderr.`,
+		Action: start,
```
nmiyake: Since your actions no longer require the CLI context, I think it's cleaner to have the `Action:` define the function and have the functions themselves be separate functions that don't take a context:

```go
Action: func(_ cli.Context) error {
   return start()
}

func start() error {
    // ...
}
```

If this function were to eventually use flags, then they should be parsed in the `Action` function and the `start` function should receive the arguments as Go types.

init/cli/status.go
```diff
@@ -28,34 +24,20 @@ func statusCommand() cli.Command {
...
-				Value: defaultPidFile},
-		},
-		Action: doStatus,
+Determines the status of the process the PID of which is written to var/run/service.pid.
+Exits:
+- 0 if the pidfile exists and can be read and the process is running
+- 1 if the pidfile exists and can be read but the process is not running
+- 3 if the pidfile does not exist or cannot be read
+If exit code is nonzero, writes an error message to stderr.`,
+		Action: status,
```
nmiyake: Same feedback as for previous command file.

init/cli/status.go
```diff
@@ -28,34 +24,20 @@ func statusCommand() cli.Command {
...
-		return respondError(msg, err, isRunning)
-	}
-
-	switch isRunning {
-	case 0:
-		pid, _ := lib.GetPid(pidFile)
-		return respondSuccess(isRunning, fmt.Sprintf("Running (%d)\n", pid))
-	case 1:
-		return respondSuccess(isRunning, "Process dead but pidfile exists\n")
+		return cli.WithExitCode(status, err)
```
nmiyake: Inline call and check:

```go
if _, status, err := lib.GetProcessStatus(); err != nil {
```

init/cli/stop.go
```diff
@@ -0,0 +1,60 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cli
+
+import (
+	"github.com/palantir/pkg/cli"
+	"github.com/palantir/go-java-launcher/init/lib"
+	"os"
```
nmiyake: Don't think this passes checks -- imports are not sorted properly. Run `./godelw format` or `./godelw verify` to run the formatting task on this file.

init/cli/stop.go
```diff
@@ -0,0 +1,60 @@
...
+	"fmt"
+)
+
+func stopCommand() cli.Command {
+	return cli.Command{
+		Name: "stop",
+		Usage: `
+Stops the process the PID of which is written to var/run/service.pid. Returns 0 if the process is successfully stopped
+or is not running and the pidfile is removed and returns 1 otherwise.`,
+		Action: stop,
```
nmiyake: Define command function here so `cli.Context` not passed to function

init/cli/stop.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+}
+
+func stop(_ cli.Context) error {
+	// The status tells us more than the error
+	process, status, _ := lib.GetProcessStatus()
+
+	switch status {
```
nmiyake: Inline:

```go
switch process, status, _ := lib.GetProcessStatus(); status {
...
```

init/cli/stop.go
```diff
@@ -0,0 +1,60 @@
...
+		return nil
+	case 1:
+		if err := os.Remove(lib.Pidfile); err != nil {
+			return cli.WithExitCode(1, err)
+		}
+		return nil
+	case 3:
+		return nil
+	default:
+		msg := fmt.Sprintf("internal error, process status code not a known value: %d", status)
```
nmiyake: Use `errrors.Errorf(fmtStr, fmtArgs)` rather than `Sprintf` + `errors.New`

init/cli/stop.go
```diff
@@ -0,0 +1,60 @@
...
+		err := errors.New(msg)
+		return cli.WithExitCode(1, err)
```
nmiyake: Inline all:

```go
return cli.WithExitCode(1, errors.Errorf("internal error, ...: %d", status))
```

init/lib/start.go
```diff
@@ -15,32 +15,26 @@
...
+	cmd.Stderr = outputFile
+	if err := cmd.Start(); err != nil {
+		return errors.Wrap(err, "failed to start command")
 	}
 
 	pid := cmd.Process.Pid
-	err = ioutil.WriteFile(pidFileName, []byte(strconv.Itoa(pid)), 0644)
-	if err != nil {
-		return pid, fmt.Errorf("Failed to write pid file: %s", pidFileName)
+	if err := ioutil.WriteFile(Pidfile, []byte(strconv.Itoa(pid)), 0644); err != nil {
```
nmiyake: Since `pid` is no longer being returned, just do `strconv.Itoa(cmd.Process.Pid)` (no need to store in separate variable)

integration_test/go_init_integration_test.go
```diff
@@ -0,0 +1,228 @@
...
+	assert.EqualError(t, err, "open var/run/service.pid: no such file or directory")
+}
+
+func TestInitStop_FailsRunningDoesNotTerminate(t *testing.T) {
+	setup()
+	defer lib.Teardown()
+
+	unstoppableCommand := "trap '' 15; /bin/echo go-init-testing && /bin/sleep 10000 &"
+	if err := exec.Command("/bin/sh", "-c", unstoppableCommand).Run(); err != nil {
+		panic(err)
```
nmiyake: In tests, I recommend storing error and using `require.NoError` to assert that error is `nil`. So, in this case:

```go
err := exec.Command("/bin/sh", "-c", unstoppableCommand).Run()
require.NoError(t, err, ...)
```

init/lib/test_util.go
```diff
@@ -0,0 +1,70 @@
...
+	"strings"
+)
+
+var files = []string{LauncherStaticFile, LauncherCustomFile, OutputFile, Pidfile}
+
+func setup() {
+	Setup("testdata/launcher-static-null.yml", "testdata/launcher-custom-null.yml")
+}
+
+func Setup(launcherStaticFilePath string, launcherCustomFilePath string) {
```
nmiyake: For test functions, recommend having them take a `t *testing.T` as the first parameter and doing a `require.NoError` rather than `panic`.

Generally, making test utilities is not recommended -- these functions aren't actually test functions because they're fully exported and callable from any other Go package (and get compiled as part of the non-test program).

init/lib/stop.go
```diff
@@ -0,0 +1,55 @@
...
+		return errors.Wrap(err, "failed to stop process")
+	}
+
+	return nil
+}
+
+func waitForProcessToStop(process *os.Process) error {
+	numSecondsToWait := 240
+	counter := 0
+	for isRunning(process) && counter < numSecondsToWait {
```
nmiyake: It's more idiomatic to use timer and ticker channels here rather than doing an explicit sleep and counter. Here's an implementation in that manner:

```go
func waitForProcessToStop(process *os.Process) error {
	waitDuration := 240 * time.Second
	timer := time.NewTimer(waitDuration)
	defer timer.Stop()
	ticker := time.NewTicker(time.Second)
	defer ticker.Stop()

	for {
		select {
		case <-ticker.C:
			if !isRunning(process) {
				return nil
			}
		case <-timer.C:
			return errors.Errorf("failed to wait for process to stop: process with pid '%d' did not stop within %v", process.Pid, waitDuration)
		}
	}
}
```

init/lib/status_test.go
```diff
@@ -14,28 +14,56 @@
...
+func TestGetProcessStatus_NotRunningPidfileDoesNotExist(t *testing.T) {
+	setup()
+	defer Teardown()
+
+	process, status, err := GetProcessStatus()
+
+	assert.Empty(t, process)
+	assert.Equal(t, status, 3)
+	msg := fmt.Sprintf("failed to read pidfile: open %s: no such file or directory", Pidfile)
+	assert.Contains(t, err.Error(), msg)
```
nmiyake: The `assert` and `require` functions take format strings, so no need to create `msg` separately:

```go
assert.Contains(t, err.Error(), "failed to read pidfile: open %s: no such file or directory", Pidfile)
```

init/lib/status.go
```diff
@@ -15,50 +15,43 @@
...
-func GetPid(pidFile string) (int, error) {
-	bytes, err := ioutil.ReadFile(pidFile)
-	if err != nil {
-		return -1, err
-	}
-	pid, err := strconv.Atoi(string(bytes[:]))
-	if err != nil {
-		return -1, err
+func isRunning(process *os.Process) bool {
+	if err := process.Signal(syscall.Signal(0)); err != nil {
```
nmiyake: Just return comparison directly:

```go
return process.Signal(syscall.Signal(0)) == nil
```

nmiyake (changes requested):
Thanks for the PR! Left some feedback.

If possible, it would be nice if we could avoid exporting "test" functions -- as currently written, would probably require moving the integration test into the same directory. I don't think that's necessarily a blocker, though, so would work on fixing the other things first.

cbrockington:
Sorry, I think my previous message was unclear, I essentially meant that testing _also_ becomes easier with the configuration options, not that testing is the only reason for them.

I think the issue is a difference in perspective of whether this is a tool which "only supports sls-specification and nothing else" or "a tool which performs it's own defined set of tasks, and also the sls-spec".  Personally, I would argue that since this is open-sourced, it has it's own API, and that API includes a bunch of optional configuration, removing that would be a back-compat breaking change which could easily break others workflows.  Maybe we / the project maintainers are fine with that, though I definitely don't think it can be argued it's not part of the API.

init/cli/start.go
gregakinman: nice, didn't know you could do this

init/cli/stop.go
gregakinman: looks like the check isn't catching it:

```
$ ./godelw format
$ echo $?
0
```

gregakinman: oh, the task does the formatting rather than check if it's formatted correctly, i see

init/cli/stop.go
gregakinman: _slick_

init/lib/test_util.go
gregakinman: > making test utilities is not recommended

You mean "making _public_ test utilities", right? As in, the problem here is that they're exported & published with the main build artifact - it should be fine to have package-private test utilities in a file that ends in `_test.go`

nmiyake: Yes exactly

integration_test/go_init_integration_test.go
gregakinman: For better readability/transparency when there's an error?

nmiyake: Correct -- `panic` statements can print in an interleaved way with tests when run. This failure mode is much easier to work with. Also has the bonus that it nests a bit less (you don't have to do an `if` check on every call that could produce an error, which is super common in tests)

gregakinman:
Agreed that this project has its own API the backwards compatibility of which should be preserved in general (and the optional flags were of course part of it). However, the code that was there before wasn't complete (though it looks like it was published in a couple releases of go-java-launcher (which makes me think maybe we should move go-init to a different repository?), I would consider the go-init part of it pre-release) so the chance somebody was using it is basically nil. We won't be breaking any workflows, so I'm fine with making the change and would prefer to.

init/cli/cli.go
iamdanfox: I don't think it's really acceptable to break workflows that people may be relying on without warning, even if they aren't in the spec. If we're converting bash to go, then we need to maintain the [previous behaviour](https://github.com/palantir/sls-packaging/blob/68497f1505f38f40560fec1015ca4871854f559f/gradle-sls-packaging/src/main/resources/init.sh#L163):

```
        echo "Usage: $0 {status|start|stop|console|restart|check${COMMANDS}}"
```

Searching internally, it seems people are relying on `init-reload.sh`.  If we want to get rid of these commands, we could do a release of sls-packaging that deprecates them (and warns people that they shouldn't be used), then do a major release later that removes them.

init/cli/stop.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/palantir/go-java-launcher/init/lib"
+)
+
+func stopCommand() cli.Command {
+	return cli.Command{
+		Name: "stop",
+		Usage: `
+Stops the process the PID of which is written to var/run/service.pid. Returns 0 if the process is successfully stopped
+or is not running and the pidfile is removed and returns 1 otherwise.`,
```
iamdanfox: Documentation should also include how long this will wait for a process to exit before it is considered 'failed to stop'

init/cli/cli.go
gregakinman: Yep, would prefer to go the route of "deprecate everything that isn't start/stop/status in sls-packaging then major version bump when we bundle this in sls-packaging instead" rather than "drop-in replacement for the current init.sh without sls-packaging major version bump"

(also obviously we will still have init.sh but it will be a simple pass-through to this binary)

iamdanfox: Ah ok I didn't know whether you wanted to delete the `init.sh` and cutover immediately to this go implementation!  Deprecated then major bump sounds good.

iamdanfox:
I'm really happy that this means we're actually gonna have tests for the init script!  Also it opens the door to do some retention on the startup log so that it can't grow in an unbounded way (cc @ellisjoe).

init/lib/status_test.go
gregakinman: hm, I'm not seeing this in practice (with the exact code you listed there):

```
--- FAIL: TestGetProcessStatus_NotRunningPidfileDoesNotExist (0.00s)
        Error Trace:    status_test.go:67
	Error:		"failed to read pidfile: open var/run/service.pid: no such file or directory" does not contain "failed to read pidfile: open %s: no such file or directory"
	Messages:	var/run/service.pid
```

I'll keep debugging, let me know if anything jumps out at you

gregakinman:
Yep! Way better than the way we tested our bash... https://github.com/palantir/sls-packaging/pull/29#issuecomment-164883865 😂

init/lib/status_test.go
gregakinman: Yeah, it doesn't look like this is the case unless I'm misreading something:

```
func Contains(t TestingT, s, contains interface{}, msgAndArgs ...interface{}) bool {
	if h, ok := t.(tHelper); ok {
		h.Helper()
	}

	ok, found := includeElement(s, contains)
...
```
I think the error message if it fails can be a format string, but not the actual `contains` since it's untyped

gregakinman:
Ok, I think all of the PR feedback was addressed, lmk if anyone has more comments. ~~Sadly still can't SSH :(~~ of course logging out and back in worked

init/lib/status_test.go
cbrockington: @gregakinman you are correct, the formatting is only for a message to produce on the assertion failing, not on the thing to check, so you were correct the first time.

gregakinman:
any merge+release objections?

nmiyake:
Sorry haven't had a chance to review the modifications -- can commit to doing so before tomorrow.

init/cli/start.go
```diff
@@ -29,59 +28,35 @@ func startCommand() cli.Command {
...
-		return respondError(msg, err, 1)
+		return cli.WithExitCode(1,
+			errors.Wrap(err, "failed to assemble command from static and custom configuration files"))
 	}
 
-	pid, err := lib.StartCommandWithOutputRedirectionAndPidFile(cmd, stdoutFile, pidfile)
-	if err != nil {
-		msg := fmt.Sprintln("Failed to start process", err)
-		return respondError(msg, err, 1)
+	if err = lib.StartCommand(cmd, outputFile); err != nil {
```
nmiyake: Minor, but when using this style, do a `err :=` assignment -- this ensures that the variable is scoped just to this block of code (and doesn't update the value of the outer `err` variable)

nmiyake (commented):
Very minor feedback, but otherwise LGTM

nmiyake:
Content looks good to me on the whole -- made a very minor comment, but shouldn't blcok merging/approval. I don't have merge permissions on the repo, though, so I can't actually do the merge.

PR #48, opened 2018-06-18
Allow for definition and launch of process groups

cbrockington:
Addition of `secondaries` configuration block can include additional launcher configurations of named processes.

Running the launcher will result in the forking and subsequent execution of secondary process, as well as a monitor for the process group, before executing the primary process.  The additional monitor process will send a kill signal to the process group after the primary process dies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,10 +35,10 @@ args:
 dirs:
   - var/data/tmp
   - var/log
-# OPTIONAL - A map of configurations of sub-processes to launch
-sub-processes:
+# OPTIONAL - A map of configurations of subProcesses to launch
+subProcesses:
   SUB_PROCESS_NAME:
-    # another StaticLauncherConfig, cannot have it's own sub-processes
+    # another StaticLauncherConfig, cannot have it's own subProcesses
     configType: executable
     configVersion: 1
     env:
@@ -68,9 +68,9 @@ dirs:
   - var/data/tmp
   - var/log
 # OPTIONAL - A map of configurations of secondary processes to launch
-sub-processes:
+subProcesses:
   SUB_PROCESS_NAME:
-    # another StaticLauncherConfig, cannot have it's own sub-processes
+    # another StaticLauncherConfig, cannot have it's own subProcesses
     configType: executable
     configVersion: 1
     env:
@@ -95,9 +95,9 @@ env:
 jvmOpts:
   - '-Xmx2g'
 # OPTIONAL - A map of configurations of secondary processes to launch
-sub-process:
+subProcess:
   SUB_PROCESS_NAME:
-    # another CustomLauncherConfig, cannot have it's own sub-processes
+    # another CustomLauncherConfig, cannot have it's own subProcesses
     configType: executable
     configVersion: 1
     env:
@@ -126,8 +126,8 @@ and `<custom.xyz>` refer to the options from the two configuration files, respec
 Note that the custom `jvmOpts` appear after the static `jvmOpts` and thus typically take precendence; the exact
 behaviour may depend on the Java distribution.
 
-If any sub-processes are defined, they will be launched as child processes of the main process, with all of these
-processes occupying their own process group.  Additionally, a monitor sub-process will be launched, which terminates
+If any subProcesses are defined, they will be launched as child processes of the main process, with all of these
+processes occupying their own process group.  Additionally, a monitor subProcess will be launched, which terminates
 the group, should the main process die.
 
 `env` block, both in static and custom configuration, supports restricted set of automatic expansions for values
--- a/README.md
+++ b/README.md
@@ -35,6 +35,10 @@ args:
 dirs:
   - var/data/tmp
   - var/log
+# OPTIONAL - A map of configurations of secondary processes to launch
+secondaries:
+  name:
+    StaticLauncherConfig
 ```
 
 ```yaml
@@ -55,6 +59,10 @@ args:
 dirs:
   - var/data/tmp
   - var/log
+# OPTIONAL - A map of configurations of secondary processes to launch
+secondaries:
+  name:
+    StaticLauncherConfig
 ```
 
 ```yaml
@@ -70,6 +78,10 @@ env:
 # Additional JVM options to be passed to the java command, will override defaults in static config. Ignored if configType is "executable"
 jvmOpts:
   - '-Xmx2g'
+# OPTIONAL - A map of configurations of secondary processes to launch
+secondaries:
+  name:
+    CustomLauncherConfig
 ```
 
 The launcher is invoked as:
--- a/README.md
+++ b/README.md
@@ -38,9 +38,8 @@ dirs:
 # OPTIONAL - A map of configurations of subProcesses to launch
 subProcesses:
   SUB_PROCESS_NAME:
-    # another StaticLauncherConfig, cannot have it's own subProcesses
+    # another StaticLauncherConfig though it cannot have its own subProcesses, and uses its parent's configVersion
     configType: executable
-    configVersion: 1
     env:
       CUSTOM_VAR: CUSTOM_VALUE
     executable: "{{CWD}}/service/lib/envoy/envoy"
@@ -70,9 +69,8 @@ dirs:
 # OPTIONAL - A map of configurations of secondary processes to launch
 subProcesses:
   SUB_PROCESS_NAME:
-    # another StaticLauncherConfig, cannot have it's own subProcesses
+    # another StaticLauncherConfig though it cannot have its own subProcesses, and uses its parent's configVersion
     configType: executable
-    configVersion: 1
     env:
       CUSTOM_VAR: CUSTOM_VALUE
     executable: "{{CWD}}/service/lib/envoy/envoy"
@@ -97,9 +95,8 @@ jvmOpts:
 # OPTIONAL - A map of configurations of secondary processes to launch
 subProcess:
   SUB_PROCESS_NAME:
-    # another CustomLauncherConfig, cannot have it's own subProcesses
+    # another CustomLauncherConfig though it cannot have its own subProcesses, and uses its parent's configVersion
     configType: executable
-    configVersion: 1
     env:
       CUSTOM_VAR: CUSTOM_VALUE
 ```
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 
 # go-java-launcher
 
-A simple Go program for launching programs from a fixed configuration. This program replaces Gradle-generated Bash
+A simple Go program for launching processes from a fixed configuration. This program replaces Gradle-generated Bash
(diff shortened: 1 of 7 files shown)


cbrockington:
@iamdanfox, @nmiyake, @gregakinman tools for "side-car" management.

Will need to be analogued in the `go-init` work in the other PR, can either do that here, after the other has merged, or in a subsequent PR.

iamdanfox:
cc infra support PoC @dansanduleac

gregakinman:
I remember discussing doing it this way (using POSIX process groups) and looking at the code from your first commit when I was in London but I remember coming away from that meeting with the impression we had decided we weren't going to do it this way. But I can't remember the reason why - do you remember?

cbrockington:
We were preferencing switching to using go-init as the long term solution, which was the side you were working on.  I figured that whilst we are switching everything over, we would still want to maintain parity between the launcher and the init, particularly given we are going to have to upgrade through deprecations.  In order to get a working environment more quickly I assumed we would do something like:

1. bump sls-packaging to include the go-java-launcher version with this PR, and deprecating non sls-spec functions in bash `init.sh` (by including this PR / some variant, we would still be able to move forward with other work as secondary processes would be supported, just via the bash `init.sh`)
2. release a major version bump with all the functionality moved into `go-init`, deprecating this
3. finally clean everything else up

If a lot of work is required to get this PR operational, then given it's short lifetime then we should give it a pass (though I had already done 90% of the work when we spoke before), but the benefit of not doing so, is we decouple having a way to launch secondary processes from chasing up people who were using non sls-spec functions previously into upgrading.

launchlib/monitor.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/pkg/errors"
+)
+
+const (
+	CheckPeriod = 5 * time.Second
+)
+
+type ProcessMonitor struct {
+	ServicePid     int
```
iamdanfox: Would prefer not to bake in assumptions about 'service' - can we just call these a PrimaryPid and ProcessGroupID ?

launchlib/monitor.go
```diff
@@ -0,0 +1,84 @@
...
+	return nil
+}
+
+func (m *ProcessMonitor) isAlive() bool {
+	process, err := os.FindProcess(m.ServicePid)
+	if err != nil {
+		return false
+	}
+
+	err = process.Signal(syscall.Signal(0))
```
iamdanfox: What does this do? Could we have a comment explaining its importance?

launcher/main/main.go
```diff
@@ -60,18 +93,64 @@ func main() {
...
-		fmt.Println("Failed to assemble executable metadata", cmd, err)
+		fmt.Println("Failed to assemble executable metadata", cmds, err)
 		panic(err)
 	}
 
-	// Execute command
-	execErr := syscall.Exec(cmd.Path, cmd.Args, cmd.Env)
+	if len(cmds.Secondaries) != 0 {
+		// Ensure we are in our own process group, since the monitor kills the group
+		if err := syscall.Setpgid(0, 0); err != nil {
```
iamdanfox: comment of what this does would be nice

launcher/main/main.go
```diff
@@ -60,18 +93,64 @@ func main() {
...
+			fmt.Println("Failed to start process monitor for service process group")
+			panic(err)
+		}
+
+		for name, secondary := range cmds.Secondaries {
+			// Ensure child processes are in the same parent group
+			if secondary.SysProcAttr == nil {
+				secondary.SysProcAttr = &syscall.SysProcAttr{}
+			}
+			secondary.SysProcAttr.Setpgid = false
```
iamdanfox: this is the crucial line to make sure subprocesses _inherit_ the process group id of the launcher

README.md
```diff
@@ -35,6 +35,10 @@ args:
 dirs:
   - var/data/tmp
   - var/log
+# OPTIONAL - A map of configurations of secondary processes to launch
+secondaries:
+  name:
+    StaticLauncherConfig
```
iamdanfox: Maybe one copy & pastable example, possibly even mentioning envoy (to help people understand the motivation)

gregakinman:
As discussed earlier today, let's nail down the config format since that will be the thing that we will have to maintain backcompat for and use in the go-init work as well. What we have in this PR looks doable to me as is, but I do have a question of if we want to designate a "primary" process vs other "secondaries." End of last week I was sketching up something similar but not exactly the same as the following:
```
type CompositeStaticLauncherConfig struct {
    LauncherConfig         `yaml:",inline"`
    StaticLauncherConfigs  []StaticLauncherConfig        `yaml:"staticLauncherConfigs"`
```
and adding `composite` or some other better name as an additional config type. Parallel format for the custom config too of course. We can unmarshal the whole thing as `LauncherConfig`, read the type, switch on it, then unmarshal the whole thing again as the stated type now that we have an indication what type it is (this is basically what we would have to do were we to introduce a new, not-backwards-compatible v2 config version but still wanted to support reading the v1, kind of like what our internal changelog parser has to do except the version is in the YAML itself and not in the filename).

So I guess the question is do we think there's value in designating the process set as primary + secondaries vs. a set of peers? From the perspective of go-init, it doesn't matter what's primary/secondary, it will just start/stop everything and report the most severe status of any process in the set. Things that need to know which process is the primary one are other things like health checkers, etc. There may be something I'm forgetting here though. Thoughts?

cbrockington:
I agree `go-init` probably won't care whether it is a list of things, or a designated primary and secondary, however, the sls service spec does have some underlying assumptions around there being a "service process" (for instance, expecting there to be one startup.log file).  Unless we decide to put all process startup logs into the same file (which I don't think we should do), it becomes unclear which of our set of peers is supposed to use the "main "startup.log".

The other reason I prefer the separation is that, like you said, other sls operations expect one process to be in charge of reporting health etc; keeping the primary/secondary distinction here I think helps to enforce that, and better declares the intent of "you have one main thing you are running, but maybe you want to run some things on the side to supplement that process".

As a side, I'm about to rename secondaries to subprocesses if that helps.

gregakinman:
Synced a bit offline, generally in agreement. Startup logs are not mentioned in the SLS spec but I agree it still makes sense for go-init to know which process to redirect to `startup.log` and which to redirect to "other" startup log files.

Kind of nitpicky but I think `secondaries` is in fact better than `subprocesses` since the latter could imply that the other processes are children of the main process in the Unix sense when they're not. Why not `sidecars` and be explicit about the use case for this?

cbrockington:
So they are currently subprocesses of the main process.

iamdanfox:
I don't really like the "secondaries" name because it sounds weird pluralized to me.  I like that there's one primary process (the real server) as this is best semantically aligned to how all our servers are deployed.  Subprocesses seems fine to me, otherwise I'd vote for `auxiliaries`.

cbrockington:
My preference is for `subprocesses` and sticking with the current configuration of `subprocesses` being children of the main process in unix itself, since this is easy to do, and follows through the operational intent down to the OS layer in the instance that someone is having to manually poke around for some reason.

If my task tracking is correct, the only remaining open concern is that currently any logs from the start-up of subprocesses just get piped to dev-null.  The launcher itself currently only knows to pipe the main process logs to `stdout`.

Perhaps the subprocesses should also just do the same?  This will result in things being interleaved, though on Linux `stdout` is at least line buffered so the interleaving will happen at the line level not character level.

gregakinman:
Ah, right, they're subprocesses here because go-java-launcher fork-execs them and then just execs the main process and becomes it. I guess the real reason I don't want to name them `subprocesses`, then, is because when launched by go-init, the main process will not be the parent of the other processes since go-init fork-execs all of them, and I just don't want to put something in the config format, which is API, that will not reflect the long-term reality.

cbrockington:
There would be some slight increase in complexity, but the isn't anything stopping go it from creating a similar process tree. Given that the go-init work is still just theory I'd rather shelf this discussion for then, after determining if it is too difficult to keep the process structure. If there are API break concerns there is nothing stopping us from holding off on a release, and the chance of someone using this unreleased API in the meantime is much less then the previous released API we broke in the last PR.

iamdanfox:
@gregakinman let's just run with `subprocesses` to avoid bikeshedding forever. It's descriptive at a high-level and accurate enough I think.

cbrockington:
Logging to stdout/stderr as well now, added a couple tests and reworked how the LauncherConfig was specified as to avoid specifying the config version more than once since it should be fixed for the whole file e.g.

NOT
```
configVersion: 1
configType: java
subProcesses:
  other:
    configVersion: 1
    configType: java
```

BUT
```
configVersion: 1
configType: java
subProcesses:
  other:
    configType: java
```

gregakinman:
it ain't just theory! :) https://github.com/palantir/go-java-launcher/tree/gk/go-init-multiprocess

README.md
```diff
@@ -35,12 +35,24 @@ args:
 dirs:
   - var/data/tmp
   - var/log
+# OPTIONAL - A map of configurations of sub-processes to launch
+sub-processes:
+  SUB_PROCESS_NAME:
+    # another StaticLauncherConfig, cannot have it's own sub-processes
```
gregakinman: its*

README.md
```diff
@@ -35,12 +35,24 @@ args:
...
+    configType: executable
+    configVersion: 1
```
gregakinman: remove to reflect update of when version is specified

README.md
```diff
@@ -35,12 +35,24 @@ args:
 dirs:
   - var/data/tmp
   - var/log
+# OPTIONAL - A map of configurations of sub-processes to launch
+sub-processes:
```
gregakinman: think you've moved this to camelCase in the code itself, update here as well

integration_test/doc.go
```diff
@@ -1,18 +0,0 @@
-// Copyright 2016 Palantir Technologies, Inc.
```
gregakinman: where did the integration tests go?

cbrockington: No idea, I obviously wasn't paying attention to a `git add -A`.

README.md
cbrockington: Thanks, not sure how Goland missed this.

README.md
cbrockington: 👍

README.md
cbrockington: 👍

dansanduleac:
Slight nit but isn't subprocess a word? Why capitalize the P?

cbrockington:
My intellij spell check was saying it should be hyphenated 🤷‍♂️

integration_test/doc.go
gregakinman: much better now! :)

gregakinman (approved):

gregakinman:
@palantir/infrastructure any nacks?

PR #49, opened 2018-07-03
go-init multi-process management

gregakinman:
Now go-init also runs the configured `subProcesses`!


init/cli/start.go
```diff
@@ -15,48 +15,36 @@
...
 func startCommand() cli.Command {
 	return cli.Command{
 		Name: "start",
 		Usage: `
-Launches the process defined by the static and custom configurations at service/bin/launcher-static.yml and
-var/conf/launcher-custom.yml. Writes its PID to var/run/service.pid and redirects its output to var/log/startup.log. If
-successful, exits 0, otherwise exits 1 and writes an error message to stderr.`,
+Ensures the service defined by the static and custom configurations at service/bin/launcher-static.yml and
+var/conf/launcher-custom.yml is running and its outputs are redirecting to var/log/startup.log and other
+var/log/$PROCESS-startup.log files. If successful, exits 0, otherwise exits 1 and writes an error message to stderr.`,
```
cbrockington: `var/log/${SUB_PROCESS}-startup.log`.

cbrockington: Also do we want to write the error just to stderr.  Writing it there alone seems like a terrible idea given that is presumably either consumed by the wrapping `init.sh` or maybe goes somewhere in the calling process? Do we not want to make sure it also gets written to the specific `startup.log` file?

init/lib/status.go
```diff
@@ -17,38 +17,89 @@ package lib
...
+// GetServiceStatus determines the status of the service based on the configuration and the pidfile.
+// Returns (ServiceStatusInfo, status, err). Possible values are:
+// - (info, 0, nil) if the pidfile exists and can be read and all processes are running
+// - (info, 1, <err>) if the pidfile exists and can be read but at least one process is not running
+// - (info, 3, <err>) if the pidfile does not exist or cannot be read
+// - (nil, 3, <err>) if the config cannot be read
+// info contains any running processes recorded in the pidfile (since these are the ones that will need to be stopped to
+// stop the service) and the commands that are not running that are defined in the configuration files (since these are
+// the ones that will need to be started to start the service).
+func GetServiceStatus() (*ServiceStatusInfo, int, error) {
```
cbrockington: The format of this function is logically similar to what already existed so I can see why it still exists, but I dislike it quite a lot. I feel like there are essentially two separate concerns which have just been jammed into the same thing (hence why it returns a 3-tuple).

I think instead it should really be two separate functions, one for the first concern of creating the `ServiceStatusInfo`, i.e. understanding which processes are running.  Errors returned from this can then be more meaningful, since "a pid file exists, but the process isn't running", isn't really an error in the function, the function has worked correctly, it's just telling you the state of the world, and perhaps similarly for the pid file not existing (compared to the pid file not being readable, or being a folder or some such, which is probably actually an error).

A separate concern is then, given this information, how is it translated to the "init status" of the service.

This also helps avoid weird looking logic in some of our functions like
```
if info, _, err := lib.GetServiceStatus(); err == nil {
   // do nothing, no error
} else if info == nil {
   // we have an error?, but we know this by looking to see if the info is nil, not actually looking at the error?
} else {
   // ignore whether we have an error and use the info
}
```

init/cli/stop.go
```diff
@@ -27,28 +25,22 @@ func stopCommand() cli.Command {
 	return cli.Command{
 		Name: "stop",
 		Usage: `
-Stops the process the PID of which is written to var/run/service.pid. Returns 0 if the process is successfully stopped
-or is not running and the pidfile is removed and returns 1 otherwise. Waits 240 seconds for the process to stop before
-considering the execution a failure.`,
+Ensures the service defined by the static and custom configurations are service/bin/launcher-static.yml and
+var/conf/launcher-custom.yml is not running. If successful, exits 0, otherwise exits 1 and writes an error message to
+stderr. Waits for at least 240 seconds for any processes to stop.`,
```
cbrockington: While I appreciate that the process itself may take slightly more than 240 seconds, we do only wait for 240 seconds at the moment, and that language is better than "at least 240 seconds" which I feel indicates it could wait longer in under some conditions (which I don't think is true).

init/cli/stop.go
```diff
@@ -27,28 +25,22 @@ func stopCommand() cli.Command {
...
 			return stop()
 		},
 	}
 }
 
 func stop() error {
-	// The status tells us more than the error
-	switch process, status, _ := lib.GetProcessStatus(); status {
+	// We take action here based on the status, not the error
+	switch info, status, _ := lib.GetServiceStatus(); status {
```
cbrockington: Following the comment on `lib.GetServiceStatus` if we change the logic there this becomes much more understandable, namely: "get info, if an actual error getting the info (like the pidfile being a folder) then error, otherwise stop all processes listed as running".

init/lib/lib_test.go
```diff
@@ -18,22 +18,33 @@ import (
...
 	"testing"
 
 	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v2"
+
+	"github.com/palantir/go-java-launcher/launchlib"
 )
 
-var files = []string{LauncherStaticFile, LauncherCustomFile, OutputFile, Pidfile}
+var files = []string{launchlib.LauncherStaticFile, launchlib.LauncherCustomFile, launchlib.OutputFileFormat, pidfile}
```
cbrockington: This probably should have been mentioned on the previous PR but this test construction has a nasty side-effect at the moment of leaving a bunch of files dirtying up the git directory if the tests are terminated early.

We should preferably avoid this.

init/lib/start.go
```diff
@@ -17,24 +17,68 @@ package lib
...
 		return errors.Wrap(err, "failed to write pidfile")
 	}
 
 	return nil
 }
+
+func pidfileExists() bool {
+	if _, err := os.Stat(pidfile); err != nil {
+		// The only piece of information from the error we care about is if the file exists.
+		return !os.IsNotExist(err)
```
cbrockington: We are potentially throwing away a huge class of errors here, which is only made up for by the fact that we will probably (?) reencounter them later.

I think the better paradigm here is to not look before you leap; this means trying to use the pidfile and handling `IsNotExist` errors if they occur when you are trying to use the file, rather than checking for existence first.  This has the added benefit of removing potentially obscure issues such as the pidfile disappearing between the `pidfileExists` and subsequent read calls; where you would get an error that the pidfile doesn't exist, even though it not existing is just supposed to indicate there are no actively running processes.

init/lib/const.go
```diff
@@ -15,8 +15,5 @@
 package lib
 
 const (
-	LauncherStaticFile = "service/bin/launcher-static.yml"
-	LauncherCustomFile = "var/conf/launcher-custom.yml"
-	OutputFile         = "var/log/startup.log"
-	Pidfile            = "var/run/service.pid"
+	pidfile = "var/run/service.pid"
```
cbrockington: This should definitely not be named service.pid if we are changing the structure of the file to be YAML. `.pid` files are strongly expected to be a single line, with the process id in. If we want to keep the file, it should continue to just contain the pid of the service (and we perhaps make others for the sub-process pids), or we should remove the file completely and create a different one if we wish to use a YAML format.

launchlib/launcher.go
```diff
@@ -33,38 +33,49 @@ const (
 	ExecPathBlackListRegex = `[^\w.\/_\-]`
 )
 
-type ServiceCommands struct {
-	Primary      *exec.Cmd
-	SubProcesses map[string]*exec.Cmd
+type ProcCmd struct {
+	Name   string
+	Cmd    *exec.Cmd
```
cbrockington: It is considered generally good practice to avoid stuttering as it often signifies a lack of clarity.  If `procCmd` is the proc cmd, then why does `procCmd.Cmd` exist?  Presumably `procCmd.Cmd` is actually the cmd, in which case what is `procCmd` if not the cmd?

launchlib/launcher.go
```diff
@@ -33,38 +33,49 @@ const (
...
+	Stdout io.Writer
 }
 
-func CompileCmdsFromConfigFiles(staticConfigFile, customConfigFile string, stdout io.Writer) (ServiceCommands, error) {
-	staticConfig, customConfig, err := GetConfigsFromFiles(staticConfigFile, customConfigFile, stdout)
+func CompileCmdsFromConfigFiles() ([]ProcCmd, error) {
```
cbrockington: If we are reducing the scope of this function then we should also change the name since whether the `Cmds` come from files is now completely hidden from the caller.

I generally dislike the reduction anyway since I feel it makes code harder to maintain as you suddenly have to be concerned with all other callers of this function if you want to change any of the things which were previously inputs and it encourages adoption of more global state (namely `PrimaryOutputFile`, `LauncherStaticFile` and `LauncherCustomFile`), but that partly depends what you consider to be the package entry point. 

I would take the `lib` package to be responsible for loading given configs and compiling commands from them, and the calling packages decide exactly where those come from.  This is more representative on the current layout as well, since the launcher expects then from the command line, compared to `go-init` which hard-codes them.

launcher/main/main.go
```diff
@@ -106,8 +106,17 @@ func main() {
 		fmt.Println("Failed to assemble executable metadata", cmds, err)
 		panic(err)
 	}
+	var primaryCmd *exec.Cmd
+	subProcesses := make(map[string]*exec.Cmd)
+	for _, cmd := range cmds {
+		if cmd.Name == "primary" {
```
cbrockington: I think we do absolutely not want to provide people the additional "functionality" of specifying their config as
```
<config>
subProcesses:
  primary:
    <other config>
```
meaning that they switch which is actually the subprocess.

A pretty good rule as well is that if you are ever making decisions based on comparisons of strings you are yourself defining, you have probably defined your API wrong, since types should satisfy almost any reason to do that.  As for why it's bad.  You now have a completely implicit (i.e. totally undefined) dependency between here and https://github.com/palantir/go-java-launcher/pull/49/commits/ae22818a42bff1f460f48c1d933f7aade6d8478c#diff-0dd5d40edbd03a1d824891bcffcce657R62  it's very unclear this dependency exists, (if you are looking from over there) or where it comes from (if you are over here); I literally had to search the whole code base for the string "primary".

This is similar to the `info + status` conflation which results in the code internally representing state via an integer (only "defined" in comments), rather than on the our "information" construct.

launchlib/const.go
```diff
@@ -0,0 +1,27 @@
...
+package launchlib
+
+import (
+	"fmt"
+)
+
+const (
+	LauncherStaticFile = "service/bin/launcher-static.yml"
+	LauncherCustomFile = "var/conf/launcher-custom.yml"
+	OutputFileFormat   = "var/log/%sstartup.log"
```
cbrockington: `var/log/%s-startup.log`?

launchlib/launcher.go
cbrockington: From reading more of the PR it seems this change was in order to remove the concept of a `primary` and `subProcesses` from our programmatic model, which as per my comment here https://github.com/palantir/go-java-launcher/pull/49#discussion_r199788187 I think is a bad idea since we intrinsically need those concepts.  Not just for the the `go-java-launcher` but also for where we store the start-up log in `go-init`.

init/lib/testdata/launcher-static-multiprocess.yml
```diff
@@ -0,0 +1,17 @@
...
+args:
+  - arg1
+subProcesses:
+  sidecar:
+    configType: java
+    mainClass: Main
+    classpath:
+      - ./testdata/
+    jvmOpts:
+      - '-Xmx4M'
```
cbrockington: newline.

integration_test/testdata/launcher-static-multiprocess.yml
```diff
@@ -0,0 +1,17 @@
...
+args:
+  - arg1
+subProcesses:
+  sidecar:
+    configType: java
+    mainClass: Main
+    classpath:
+      - ./testdata/
+    jvmOpts:
+      - '-Xmx4M'
```
cbrockington: newline

init/lib/status.go
```diff
@@ -17,38 +17,89 @@ package lib
...
-	// Docs say FindProcess always succeeds on Unix.
-	process, _ := os.FindProcess(pid)
-	if !isRunning(process) {
-		return process, 1, errors.New("pidfile exists but process is not running")
+	// What processes are running (regardless of if they are configured), and which of the configured processes are not
+	// listed in the pidfile or are not running?
+	// Look at the pidfile and record what's running.
+	runningProcs := make([]*os.Process, 0, len(servicePids.PidsByName))
+	for _, procPid := range servicePids.PidsByName {
+		// Docs say FindProcess always succeeds on Unix.
```
cbrockington: Might be worth creating `isPidRunning` since we do this in more than one place and now both places should document the weird "FindProcess always succeeds" behaviour.

init/lib/start.go
```diff
@@ -17,24 +17,68 @@ package lib
...
+	procCmd.Cmd.Stdout = procCmd.Stdout
+	procCmd.Cmd.Stderr = procCmd.Stdout
+	if err := procCmd.Cmd.Start(); err != nil {
 		return errors.Wrap(err, "failed to start command")
 	}
+	return nil
+}
 
-	if err := ioutil.WriteFile(Pidfile, []byte(strconv.Itoa(cmd.Process.Pid)), 0644); err != nil {
+func writePid(name string, pid int) error {
```
cbrockington: It feels quite gross to me that we write pids by:
1. have the name and pid
2. write the name and pid
3. read the name and pid from where we just wrote it
4. add a new name and pid to the ones we have just read
5. write both names and pids

launchlib/launcher.go
```diff
@@ -33,38 +33,49 @@ const (
 	ExecPathBlackListRegex = `[^\w.\/_\-]`
 )
 
-type ServiceCommands struct {
-	Primary      *exec.Cmd
-	SubProcesses map[string]*exec.Cmd
+type ProcCmd struct {
+	Name   string
+	Cmd    *exec.Cmd
+	Stdout io.Writer
```
cbrockington: This should probably not be here.  It is pretty bad practice to be returning things which need to be closed from anything but minimal helper functions as it immediately becomes very unclear who is responsible for closing them.  At the moment, it doesn't look like they get closed anywhere.

cbrockington (changes requested):
I've proposed quite a few changes so haven't reviewed the tests yet (other than the fact it would be nice if they didn't dirty the git directory), since I imagine they will change relatively substantially with all the fixes.

init/cli/start.go
gregakinman: Good thought - when we write things to `stderr` we're at the level of abstraction where it's not clear which `startup.log` file to write it to. So all we really can do is write to `stderr`. The client can do with that what they wish. In the main use case for this, `init.sh`, it probably makes the most sense for `init.sh` to redirect it to its own `stderr`, which the deployment infrastructure should read (since it executed `init.sh`) and deal with accordingly.

init/cli/stop.go
gregakinman: This is just me being really pedantic. The way we wait 240 seconds is by creating a new `Timer`, and the docs for `NewTimer` say:
```
// NewTimer creates a new Timer that will send
// the current time on its channel after at least duration d.
```

init/lib/lib_test.go
gregakinman: This is why I `defer teardown()` so that if something panics then this stuff gets cleaned up. If one terminates the program in such a way that deferred functions are not run, it's reasonable to expect that certain things the program was doing might not be cleaned up.

init/lib/start.go
gregakinman: Intentionally throwing away a huge class of errors here because we don't care about them here. All this function does is say if the file exists - it isn't designed to notify the caller of anything else. Encountering an error such as "the file does exist but is somehow bogus" is the responsibility of the caller to handle (and is handled).

However, your note about the pidfile disappearing between the invocation of this function and a possible subsequent read call is a good flag and is sufficient reason to not look before leaping.

launchlib/const.go
gregakinman: If we include the dash, then the main process's log file gets named `var/log/-startup.log`. Note how the subprocess log files get named `fmt.Sprintf(OutputFileFormat, "name-")`. I can see how the dash in the `Sprintf` could be confusing, though, could make a small wrapper function for this?

init/lib/start.go
gregakinman: Actually, I started to edit `writePid` to be `writePids` and take a `map[string]int` and do it in batch, but since we can't atomically start a batch of commands I had to defer a function that would write all the pids of commands started so far (in case one of the later commands failed to start) so that we would still record which processes are now running. That feels to me grosser than what we already have. Maybe I'm wrong and deferring like that is considered less gross in the golang world? I'm not really sure what call to make here.

launchlib/launcher.go
gregakinman: Yikes, thanks for catching this.

init/cli/start.go
gregakinman: Actually, I see us already logging things during the config parsing process to the primary startup log file, and those things may not necessarily be just about the primary process. So looks like there's precedent for doing that. I'll change the cli exit functionality to log errors to the primary startup log as well as to stderr.

launchlib/launcher.go
gregakinman: Ok, I'm changing everything to not return open files so it's clear who owns their closing, but in doing so, I'm seeing that it seems like we can't even close these files ourselves anyway. Each is opened right before we set each `Cmd`'s `Stdout` to it. Then our program starts the `Cmd` and then exits and the child process gets adopted by `init`. The process image is completely replaced by whatever command we started. It seems the only way to ensure these files' closure is to let the OS do it when the process terminates. I assume this is doing at the OS level exactly what bash does with redirects?

launcher/main/main.go
gregakinman: I agree this is smelly, and I knew that when I wrote it -  here was my thought process behind this decision: `go-java-launcher` needs to know which command should be run as the primary process, but `go-init` doesn't (since each process is bundled with its output filename). The return value from `CompileCmdsFromConfigFiles` (the struct) will conform to one of those use cases better than it will the other, and the other use case will have to do something suboptimal to account for that. (`go-java-launcher` has to do what is written here; `go-init` has to do something that is, admittedly, substantially less suboptimal, but is still a bit weird - it has to apply the same logic to an array of a type as well as just a single value of that type, whereas it would be cleaner if it just received an array that also included that single value).

Given the choice of which use case I wanted to tailor this to, I chose `go-init`, since that's our long-term solution, and `go-java-launcher` launching multiple processes is a short-term solution. In hindsight this was kind of dumb and I'm being way too much of a perfectionist here. As part of moving `CompileCmdsFromConfigFiles` to the `go-init` side I'll change this back to the typed version.

launchlib/launcher.go
gregakinman: > I feel it makes code harder to maintain as you suddenly have to be concerned with all other callers of this function if you want to change any of the things which were previously inputs

I understand the mentality behind this, but in this case, I don't think the additional maintainability cost is that high.

>  it encourages adoption of more global state

I see it less as relying on global state and more as intentional lack of configurability.

> that partly depends what you consider to be the package entry point.

Indeed. I see that `CompileCmdsFromConfigFiles` is only used by `go-init`, so I'll just move it to `go-init` and also add the adapter from struct-that-identifies-primary-process to array-of-equal-processes there.

launchlib/launcher.go
gregakinman: I still need this type on the `go-init` side and I can't for the life of me figure out a name that doesn't have `cmd` in it. I just need something that bundles an `*exec.Cmd` and a `string`. Going with `CmdWithOutputFile` for now...any better suggestions?

gregakinman:
I've addressed some of the feedback but ran out of time to address all of it today - I'll resume again Thursday.

init/lib/lib_test.go
cbrockington: I do not think that is reasonable, particularly given one way of it happening is just Crtl+C'ing the test run.

init/lib/start.go
cbrockington: I think you were right with the deferring method.  Deferring isn't seen in any particularly negative way and this use case seems similar to clean-up actions for which defer is recommended.

launchlib/const.go
cbrockington: I think the format should specifically be for secondary processes and a separate variable should explicitly declare the main processes output file, since it's static.

launchlib/launcher.go
cbrockington: You can close them after starting the command.  Under the hood, starting the command forks the current process which creates a copy of the parents file descriptors, we in the parent can then close the file, since the child already has it's own open copy.

launchlib/launcher.go
cbrockington: Perhaps we have a different view of the separation of concerns for different parts of this project.

I consider `launcherlib` to be a library which essentially passes two config files and creates `*exec.Cmds` representing that configuration.  As such, the "intentional lack of configurability" is more "a removal of library functionality", and it makes the library unusable by one of the two things that use it `go-java-launcher`

I'm completely fine with the idea of having a deliberate lack of configurability in `go-init`.

gregakinman: No, we're in agreement on the respective roles of the different parts of this project. `CompileCmdsFromConfigFiles` is not used by `go-java-launcher`, only by `go-init`, so it wasn't necessary for it to be general enough to take the paths to the config files as arguments. Which means it doesn't really belong in `launchlib` anymore - so I moved it to `init`.

init/lib/lib_test.go
gregakinman: Ok, installed signal handlers so that things will get cleaned up on ctrl+c (among other signals).

launchlib/launcher.go
gregakinman: Ah, right, I forgot that (I was wondering how the parent could close the file and allow the child to still read it).

This would be a good `./godelw verify` check!

init/lib/lib_test.go
cbrockington: thanks

gregakinman:
Ok, think I've addressed all the feedback so far, should be ready for another round

init/lib/status.go
```diff
@@ -17,38 +17,111 @@ package lib
...
+		primaryStdout)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to read static and custom configuration files")
+	}
+	serviceCmds, err := launchlib.CompileCmdsFromConfig(&staticConfig, &customConfig, primaryStdout)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to compile commands from static and custom configurations")
+	}
+	cmdsByName = make(map[string]CmdWithOutputFile)
+	cmdsByName["primary"] = CmdWithOutputFile{
```
cbrockington: This is still suffering from the preconceived notion that "all processes should be treated exactly the same" when we do not in fact treat them the same.  The config defines them as different, the log output defines them as different, and we want users to think of them as being different.  I don't understand why we are then constantly trying to shoehorn that into this model of everything being treated the same?

Not only does it introduce code smells (of still using a string to define a concept), but it is fundamentally buggy, since I can now define a valid configuration which will not work with `go-init`.

cbrockington: This is still suffering from the preconceived notion that "all processes should be treated exactly the same" when we do not in fact treat them the same. The config defines them as different, the log output defines them as different, and we want users to think of them as being different. I don't understand why we are then constantly trying to shoehorn that into this model of everything being treated the same?

Not only does it introduce code smells (of still using a string to define a concept), but it is fundamentally buggy, since I can now define a valid configuration which will not work with `go-init`.

init/cli/stop.go
```diff
@@ -27,34 +27,27 @@ func stopCommand() cli.Command {
...
-	case 0:
-		if err := lib.StopProcess(process); err != nil {
-			return cli.WithExitCode(1, err)
-		}
-		if err := os.Remove(lib.Pidfile); err != nil {
-			return cli.WithExitCode(1, err)
-		}
+	runningProcsByName, err := lib.GetRunningProcsByName()
+	if err != nil {
+		// If we can't determine what's running, there's nothing to stop, really.
```
cbrockington: I feel this comment doesn't make a lot of sense.  If the only possible errors returned by `lib.GetRunningProcsByName` simply signify that there are no running processes, it should just return an empty slice.  I'm pretty certain that this is not the case though.  There are loads of errors, including pretty basic things like broken configuration / pid files which could cause this to error and which definitely do not mean "everything is fine, we just have no process to stop".

init/cli/stop.go
cbrockington: Ah ok.  I didn't know that, this makes sense in that case.

init/cli/errors.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/palantir/go-java-launcher/launchlib"
+)
+
+func logErrorAndReturnWithExitCode(err error, exitCode int) cli.ExitCoder {
+	// If there's an error logging the error to the primary output file, we still want to write the error to stderr.
+	_ = logToPrimaryOutputFile(err.Error())
+	return cli.WithExitCode(exitCode, err)
+}
+
+func logToPrimaryOutputFile(log string) (rErr error) {
```
cbrockington: This is only used in the function above, and as such, can probably just take an error as the paramter then you can directly write that using `fmt.Fprintln(io.writer, interface{}...)`

init/cli/start.go
```diff
@@ -15,48 +15,37 @@
...
+var/conf/launcher-custom.yml is running and its outputs are redirecting to var/log/startup.log and other
+var/log/${SUB_PROCESS}-startup.log files. If successful, exits 0, otherwise exits 1 and writes an error message to
+stderr.`,
 		Action: func(_ cli.Context) error {
 			return start()
 		},
 	}
 }
 
 func start() error {
```
cbrockington: I find the new logical flow of this function overall very convoluted.  We currently:
- try and load the launcher config then the pid file to determine processes which aren't running
- if that fails, reload the launcher config and return everything
- start not running services
- load the pid file again
- add the new pids to the deserialized pid file
- save the pid file

Why are we:
- Load config
- Load pid file to determine which of config are already running
- Start the rest
- Save pids we knew were previously running + the ones we started, back to the pid file

This seems much more direct, and avoids constantly reserializing files from disk.

init/lib/stop.go
```diff
@@ -23,36 +23,57 @@ import (
 	"github.com/pkg/errors"
 )
 
-func StopProcess(process *os.Process) error {
-	if err := process.Signal(syscall.SIGTERM); err != nil {
-		if !strings.Contains(err.Error(), "os: process already finished") {
-			return errors.Wrap(err, "failed to stop process")
+func StopService(procs []*os.Process) error {
```
cbrockington: Keeping the names here both means we don't need to create a slice from our named process map, and also means we can be more specific in our logging than just "a" process failed to stop

launchlib/launcher.go
```diff
@@ -33,41 +33,53 @@ const (
...
+	SubProcs            map[string]*exec.Cmd
+	SubProcsOutputFiles map[string]string
 }
 
-func CompileCmdsFromConfigFiles(staticConfigFile, customConfigFile string, stdout io.Writer) (ServiceCommands, error) {
-	staticConfig, customConfig, err := GetConfigsFromFiles(staticConfigFile, customConfigFile, stdout)
-	if err != nil {
-		return ServiceCommands{}, err
+func CompileCmdsFromConfig(staticConfig *PrimaryStaticLauncherConfig, customConfig *PrimaryCustomLauncherConfig,
+	primaryStdout io.Writer) (serviceCmds *ServiceCmds, rErr error) {
```
cbrockington: Can we replace the multiple instances of calling the parameter `primaryStdout` with just `stdout`.  I think it's pretty confusing given we then create primary output files which themselves have nothing to do with the `stdout`.

launchlib/launcher.go
```diff
@@ -33,41 +33,53 @@ const (
...
 )
 
-type ServiceCommands struct {
-	Primary      *exec.Cmd
-	SubProcesses map[string]*exec.Cmd
+type ServiceCmds struct {
+	Primary             *exec.Cmd
+	PrimaryOutputFile   string
+	SubProcs            map[string]*exec.Cmd
+	SubProcsOutputFiles map[string]string
```
cbrockington: I don't think these output file definitions belong here.  One flag is the fact that elsewhere in the code we now have an "impossible edge case" which can only be hit if other code is buggy but has no direct compilation tie to the construction.

We could avoid this by making new `ServiceCmds` structures through a `NewServiceCmds` method, and additionally have the structure itself responsible for starting commands (so we can ensure ahead of time everything is set), but then that explicitly isn't how `go-java-launcher` starts things. 

As such, to me, it seems to make more sense to define this as part of the `go-init` logic.

launchlib/const.go
```diff
@@ -0,0 +1,26 @@
...
+
+package launchlib
+
+import (
+	"fmt"
+)
+
+const (
+	OutputFileName   = "startup.log"
+	OutputFileFormat = "var/log/%s" + OutputFileName
```
cbrockington: This is still not actually the `OutputFileFormat`, the output file format is `var/log/%s-startup.log`.  The `PrimaryOutputFile` is then different, and doesn't actually conform to this format.

cbrockington (changes requested):
There are still a few things going on I disagree with, such as loading config / pid files multiple times during operations, and still using the string "primary" to refer to the primary process, rather than treating it as something slightly different (which it is).

The other thing I spotted was we seem to open the primary output file in a lot of places, and it might be better to either pass this around, or set it as the process stdout right at the start.

Still have a few files to review, but heading out for the evening so submitting these know so you have something to go over before tomorrow.

launchlib/const.go
gregakinman: Can you clarify what you're getting at here?

launchlib/launcher.go
gregakinman: > elsewhere in the code we now have an "impossible edge case" which can only be hit if other code is buggy but has no direct compilation tie to the construction

Can you elaborate here? I'm not sure what you're getting at, I can't figure out what this edge case is

launchlib/const.go
cbrockington: The main process start up log gets written to `var/log/startup.log`, the format of sub-processes is then `var/log/%s-startup.log`.  I don't think the format should be `var/log/%sstartup.log` where the parameter is not the name of the sub-process, but actually `name + "-"` since it's very non-obvious the dash would be added later.

If you really want to tie these together, then I see no problem with having something like:
```
logDir = filepath.join("var", "log")
startupLogName = "startup.log"
PrimaryOutputPath = filepath.join(logDir, startupLogName)
SubProcessOuputPathFormat = filepath.join(logDir, "%s-" + startupLogName)
```

launchlib/launcher.go
cbrockington: https://github.com/palantir/go-java-launcher/pull/49/files#diff-9c3962318b8e21e39835af6f2183c281R108

gregakinman:
Have made some changes to the actual code to address the feedback - tests still WIP though.

gregakinman:
Ugh, anyone know why https://github.com/palantir/go-java-launcher/pull/49/files#diff-29944324a3cbf9f4bd0162dfe3975d88R47 is not increasing the Circle timeout for that command like the docs (https://circleci.com/docs/1.0/configuration/#modifiers) say it should? Unfortunately we have to run all the tests that wait 240s to stop serially (since they need exclusive use of the process table)... if we can't figure out how to bump the timeout, we can either cut the number of those tests, or break them into separate Circle commands or nodes, or print stuff after each sub-test so that the Circle timeout resets...

Regardless, think this is ready for more review... I reorganized the integration tests to make it more clear what situations were being tested. Right now it exhaustively tests every case, but if we think that's too verbose, we can remove some of them and get an orthogonal arrays or all-pairs kind of setup going.

README.md
```diff
@@ -136,5 +136,15 @@ for `CUSTOM_PATH`. The following fixed expansions are supported:
...
 environment cannot be overwritten with this mechanism; use the `javaHome` mechanism in `StaticLauncherConfig` instead.
 
+# go-init
+
+This repository also publishes a binary called `go-init` that supports the commands `start`, `status`, and `stop`, in
+adherence with the
+[Linux Standard Base](http://refspecs.linuxbase.org/LSB_3.1.1/LSB-Core-generic/LSB-Core-generic/iniscrptact.html)
+specification for init scripts. The binary reads configuration from (relative to its working directory)
+`service/bin/launcher-static.yml` and `var/conf/launcher-custom.yml` in the same vein as `go-java-launcher`, and outputs
+to `var/log/startup.log` and other `var/log/${SUB_PROCESS}-startup.log` files. `go-init` does not launch each
```
cbrockington, line 146: Doesn't this now output to `var/log/${SERVICE_NAME}-startup.log`?

init/cli/lib.go
```diff
@@ -0,0 +1,126 @@
...
+
+var (
+	launcherStaticFile = "service/bin/launcher-static.yml"
+	launcherCustomFile = "var/conf/launcher-custom.yml"
+	pidfile            = "var/run/pids.yml"
+)
+
+type servicePids struct {
+	Pids map[string]int `yaml:"pids"`
+}
```
cbrockington: If this doesn't have anything else, and is unlikely to ever have anything else, there is potential for making it a type alias
via `type servicePids map[string]int`.  This would slightly change the yaml structure as well though, since it would now be a straight map, not under a "pids" key, but I think that's fine.

init/cli/lib.go
```diff
@@ -0,0 +1,126 @@
...
+
+type serviceStatus struct {
+	notRunningCmds map[string]launchlib.CmdWithOutputFileName
+	writtenPids    map[string]int
+	runningProcs   map[string]*os.Process
+}
```
cbrockington: Looks good, I think it helps make if obvious what things we are interested in, and keeping them separated also makes it clear that they may not map directly too each other (e.g. a written pid, may not be running).

cbrockington: There is a potential OO argument for making a struct of essentially all of the map values for a given key, and having methods on that, but I think that's somewhat stylistic and it would be a bit of a pain to construct.  (As an example, for clarity, not an expectation of change:
```
type ServiceProcess struct {
    writtenPid int
    runningProc *os.Process
    startCmd launchlib.CmdWithOutputFileName
}

type (p *ServiceProcess) Start() error {
    if p.runningProcess != nil {
        return nil
    }
    // ... else start process
}

type ServiceStatus map[string]*ServiceProcess
```

init/cli/lib.go
```diff
@@ -0,0 +1,126 @@
...
+		return map[string]int{}, map[string]*os.Process{}, nil
+	}
+	var servicePids servicePids
+	if err := yaml.Unmarshal(pidfileBytes, &servicePids); err != nil {
+		return nil, nil, errors.Wrap(err, "failed to deserialize pidfile")
+	}
+	runningProcs := make(map[string]*os.Process)
+	for name, pid := range servicePids.Pids {
+		running, proc := isPidRunning(pid)
+		if running {
```
cbrockington:
```
if running, proc := isPidRunning(pid); running {
    runningProcs[name] = proc
}
```

init/cli/lib.go
cbrockington: This would also give us a nicer explicit type we can use for `writtenPids`

init/cli/lib.go
```diff
@@ -0,0 +1,126 @@
...
+			runningProcs[name] = proc
+		}
+	}
+	return servicePids.Pids, runningProcs, nil
+}
+
+func getConfiguredCommands(ctx cli.Context) (cmds map[string]launchlib.CmdWithOutputFileName, rErr error) {
```
cbrockington: I think the return values shouldn't be named here since we don't really need them to be (`rErr` is never used, and `cmd` is initialized with `make`)

init/cli/status.go
```diff
@@ -15,30 +15,61 @@
...
-		return cli.WithExitCode(status, err)
+func status(ctx cli.Context) (rErr error) {
+	outputFile, err := os.OpenFile(launchlib.PrimaryOutputFile, outputFileFlag, outputFileMode)
+	if err != nil {
+		return cli.WithExitCode(4, errors.Errorf("failed to create primary output file: %s",
+			launchlib.PrimaryOutputFile))
+	}
+	defer func() {
+		if cErr := outputFile.Close(); rErr == nil && cErr != nil {
+			rErr = cli.WithExitCode(4, errors.Errorf("failed to close primary output file: %s",
```
cbrockington: There is some argument to be made that logging is secondary to the function working, and that if everything would otherwise be fine, it's better for things to keep working / falling back to just stdout logging, rather than break because they can't log (obviously this would be different if it was an audit log).

init/cli/start.go
```diff
@@ -15,48 +15,99 @@
...
 	}
+	return pids, nil
+}
 
-	if err := lib.StartCommand(cmd.Primary, outputFile); err != nil {
-		return cli.WithExitCode(1, errors.Wrap(err, "failed to start process"))
+func startCommand(cmd launchlib.CmdWithOutputFileName) (rErr error) {
+	stdout, err := os.OpenFile(cmd.OutputFileName, outputFileFlag, outputFileMode)
+	if err != nil {
+		return errors.Wrap(err, "failed to open output file: "+cmd.OutputFileName)
```
cbrockington: `errors.Wrapf(err, "... %s", cmd.OutputFileName)` rather than string concat.

init/cli/start.go
```diff
@@ -15,48 +15,99 @@
...
+	}
+	defer func() {
+		if cErr := stdout.Close(); rErr == nil && cErr != nil {
+			rErr = errors.Wrap(err, "failed to close output file: "+cmd.OutputFileName)
```
cbrockington: I think this is probably something we want to log, but again I'm not sure if we should completely error.  There is unfortunately the very real argument that unless we return an error code, no one will look at the errors in the log, but I think exit codes representing whether or not the actions on the service worked tie more closing into what people care about (which "should" get cleaned up but the OS).  @iamdanfox for perhaps additional comment / consistency with other infra decisions.

init/cli/status.go
```diff
@@ -15,30 +15,61 @@
...
+
+	serviceStatus, err := getServiceStatus(ctx)
+	if err != nil {
+		return logErrorAndReturnWithExitCode(ctx, errors.Wrap(err, "failed to determine service status"), 4)
+	}
+	if len(serviceStatus.notRunningCmds) > 0 {
+		notRunningCmdNames := make([]string, 0, len(serviceStatus.notRunningCmds))
+		for name := range serviceStatus.notRunningCmds {
+			notRunningCmdNames = append(notRunningCmdNames, name)
+		}
```
cbrockington, line 45: I really wish these 4 lines weren't necessary :(

init/cli/stop.go
```diff
@@ -16,45 +16,95 @@ package cli
...
+	}
+	defer func() {
+		if cErr := outputFile.Close(); rErr == nil && cErr != nil {
+			rErr = cli.WithExitCode(1, errors.Errorf("failed to close primary output file: %s",
+				launchlib.PrimaryOutputFile))
 		}
-		if err := os.Remove(lib.Pidfile); err != nil {
-			return cli.WithExitCode(1, err)
+	}()
+	ctx.App.Stdout = outputFile
```
cbrockington: There may be a way of abstracting out this log file setup into a common function since we are doing it every time, something like:
```
func actionWithStdoutLog(action func(cli.Context) error) (func(cli.Context) error) {
    return func(ctx cli.Context) {
        // setup log
        // do defer
        action
        // potentially even do the `logError` part of `logErrorAndReturnWithExitCode`
    }
}
```

launchlib/const.go
```diff
@@ -12,11 +12,18 @@
...
-	LauncherCustomFile = "var/conf/launcher-custom.yml"
-	OutputFile         = "var/log/startup.log"
-	Pidfile            = "var/run/service.pid"
+	outputLogFile = "startup.log"
+)
+
+var (
+	logDir                     = "var/log"
+	PrimaryOutputFile          = filepath.Join(logDir, outputLogFile)
+	SubProcessOutputFileFormat = filepath.Join(logDir, "%s-"+outputLogFile)
```
cbrockington, line 28: I thought we discussed that the format is actually consistent for all of them, just the Primary uses the `ServiceName` rather than the `subProcessName`?

launchlib/launcher.go
```diff
@@ -33,44 +33,59 @@ const (
...
-func CompileCmdsFromConfig(staticConfig *PrimaryStaticLauncherConfig, customConfig *PrimaryCustomLauncherConfig, stdout io.Writer) (ServiceCommands, error) {
-	cmds := ServiceCommands{
-		SubProcesses: make(map[string]*exec.Cmd),
+func CompileCmdsFromConfig(staticConfig *PrimaryStaticLauncherConfig, customConfig *PrimaryCustomLauncherConfig,
+	stdout io.Writer) (serviceCmds *ServiceCmds, rErr error) {
+	serviceCmds = &ServiceCmds{
+		SubProcs: make(map[string]CmdWithOutputFileName),
 	}
+
 	var err error
```
cbrockington: Remove

launchlib/launcher.go
```diff
@@ -33,44 +33,59 @@ const (
...
+			if cErr := subProcStdout.Close(); rErr == nil && cErr != nil {
+				rErr = errors.Wrapf(err, "failed to close subProcess output file: %s",
+					subProcOutputFileName)
+			}
+		}()
+		subProc, err := compileCmdFromConfig(&subProcStatic, &subProcCustom, subProcStdout)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to compile command for subProcess config '%s'", name)
 		}
+		serviceCmds.SubProcs[name] = CmdWithOutputFileName{Cmd: subProc, OutputFileName: subProcOutputFileName}
```
cbrockington: Could we split the subprocess compilation part into it's own function?  Since we are also opening and closing the log file it makes this for loop pretty lengthy.

cbrockington (changes requested):
Couple of minor things but overall looks good.

One additional thing which would probably be good to change (but I can't comment on the line), is removing `ServiceName` from the `StaticLauncherConfig` and putting it in the `PrimaryStaticLauncherConfig` since it only makes sense on the top level.

README.md
gregakinman: Yeah, per our conversation a few days ago, ideally, yes, but afterwards when I looked into it I noticed `ServiceName` is unfortunately not a required field in the config, and sls-packaging doesn't fill it in. I'd rather have the file always be `startup.log` rather than sometimes `my-awesome-service-startup.log` and sometimes `-startup.log`. And making it a required field would require a major version bump :/.

`ServiceName` is luckily still usable as a string key for the primary process in the pidfile and elsewhere in the code, though, since it's just used as a unique key and not read by humans.

init/cli/lib.go
gregakinman: ooh forgot you could do this, I like this better

init/cli/stop.go
gregakinman: nice, this is what i wanted to do in my head but couldn't figure out how.

init/cli/start.go
gregakinman: Agreed. Although it's kind of goofy to log to a file "hey I couldn't close this file...obviously...because I just wrote to it..." but seems like the best thing to do.

edit: we probably want to write this to `startup.log` actually since that's where the "cli logic" logging goes

init/cli/status.go
gregakinman: yeah go doesn't really want a `.keys` method apparently ¯\_(ツ)_/¯

launchlib/const.go
gregakinman: same as https://github.com/palantir/go-java-launcher/pull/49#discussion_r202416298

'
```diff
@@ -0,0 +1,55 @@
+machine:
```
cbrockington: Is this file supposed to be called just `'`?

README.md
cbrockington: Hmmmm.  So we can enforce it to be written / exist somewhere in `sls-packaging` since we have to update it to use `go-init` anyway.

For now I'm fine with us just being weary of potential breaks from things consuming the `startup.log` from `${SERVICE_NAME}-startup.log`.  @iamdanfox you have more oversight on infra in general, anything you know of that this will break?

init/cli/cli.go
```diff
@@ -15,14 +15,52 @@
...
+	app.Subcommands = []cli.Command{startCliCommand, statusCliCommand, stopCliCommand}
 	return app
 }
+
+func executeWithContext(action func(cli.Context) error) func(cli.Context) error {
+	return func(ctx cli.Context) (rErr error) {
+		outputFile, err := os.OpenFile(launchlib.PrimaryOutputFile, outputFileFlag, outputFileMode)
+		if err != nil {
+			// Logging is secondary to actually starting the service, so just log to stdout and move on.
+			ctx.App.Stdout = os.Stdout
```
cbrockington: Isn't this the default?  I think we can just remove this, and switch it to:
```
// Fallback to default behaviour if error opening log file
if outputFile, err := os.OpenFile(....); err == nil {
    defer func() { ... }()
    ctx.App.Stdout = outputFile
    return action(ctx)
}
```

init/cli/lib.go
```diff
@@ -0,0 +1,116 @@
...
+
+	"github.com/palantir/pkg/cli"
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v2"
+
+	"github.com/palantir/go-java-launcher/launchlib"
+)
+
+const (
+	outputFileFlag = os.O_WRONLY | os.O_APPEND | os.O_CREATE
```
cbrockington: I think `APPEND`ing is a behaviour change, as I think currently if you restart a process the startup.log gets replaced.  I'm if we are APPENDING instead now, we need to have some kind of enforced retention.

init/cli/cli.go
```diff
@@ -15,14 +15,52 @@
...
 	app := cli.NewApp()
 	app.Name = "go-init"
 	app.Usage = "A simple init.sh-style service launcher CLI."
 
-	app.Subcommands = []cli.Command{startCommand(), statusCommand(), stopCommand()}
+	app.Subcommands = []cli.Command{startCliCommand, statusCliCommand, stopCliCommand}
 	return app
 }
+
+func executeWithContext(action func(cli.Context) error) func(cli.Context) error {
```
cbrockington: So.... after saying this should be extracted out (which I still think it should be) I also realised that at the moment we have a kind of weird situation where `stop` and `status` commands log to `startup.log` as well, which is slightly odd, given they aren't starting anything.

Ultimately I'm not sure whether we want to have a `launcher.log` which is used for all logging `go-init` itself would do, or something else.

integration_test/testdata/launcher-static.yml
```diff
@@ -1,5 +1,6 @@
 configType: java
 configVersion: 1
+serviceName: primary
```
cbrockington, line 3: If this isn't a requirement, then in some ways I would prefer we have a test where it is blank, since that's more likely to have issues.

cbrockington (changes requested):
Realised there are still potential logging issues from a mixture of requiring retention and non-startup things logging to startup.log

'
gregakinman: lol. git fail.

README.md
gregakinman: By "I'm fine with us just being weary of potential breaks from things consuming the `startup.log` from `${SERVICE_NAME}-startup.log`" are you advocating for changing this to actually output to `$(SERVICE_NAME}-startup.log` and then updating `sls-packaging` to fill in `ServiceName` in the config, eating the possibility that sometimes something might get logged as `-startup.log`? Or are you saying you're fine with it just always being `startup.log`?

init/cli/cli.go
gregakinman: Yeah, it's weird... I'm not sure what to do either. I think this is something we could change later if we decide we want to do something different, though.

integration_test/testdata/launcher-static.yml
gregakinman: This belongs in the config parsing testing - since what we want to test is that if we leave it off, we still get a valid string back from the parsing. Added some assertions to the config testing for this.

README.md
cbrockington: I was saying that I'm fine with the latter, though input from `infra` people would be nice.

cbrockington:
OK, no update from `infra` so I think what we should probably do for now is:
 - [X] Leave things as they are with logging the primary process to `startup.log`
 - [x] Override the startup logs when starting (less likely infinite log growth)
 - [x] Append in all other cases (getting status, stopping etc) since overriding here would be kind of sad given they don't log much, and getting the status of your service would wipe the logs of how it started, which would be bad.

Sound good?

gregakinman:
Yep, sounds good! It was already updated to have that behavior, but I just added some tests that verify it does that.

cbrockington (approved):

README.md
dansanduleac: I'm fine with leaving this as `startup.log`.
Btw, why would the service name be defined by sls-packaging?
Sls-packaging defines the *product* name, but you could be running multiple instances of the same product with different service names, so it seems like a deployment concern to set this.

circle.yml
```diff
@@ -41,7 +43,8 @@ test:
     - mkdir -p "$CIRCLE_TEST_REPORTS/$CIRCLE_PROJECT_REPONAME"
     - cd "$GO_PROJECT_SRC_PATH" && ./godelw verify --apply=false --tags none --junit-output="$CIRCLE_TEST_REPORTS/$CIRCLE_PROJECT_REPONAME/$CIRCLE_PROJECT_REPONAME-tests.xml"
     - mkdir -p "$CIRCLE_TEST_REPORTS/$CIRCLE_PROJECT_REPONAME-integration"
-    - cd "$GO_PROJECT_SRC_PATH" && ./godelw test --tags=integration --junit-output="$CIRCLE_TEST_REPORTS/$CIRCLE_PROJECT_REPONAME/$CIRCLE_PROJECT_REPONAME-integration.xml"
+    - cd "$GO_PROJECT_SRC_PATH" && ./godelw test --tags=integration --junit-output="$CIRCLE_TEST_REPORTS/$CIRCLE_PROJECT_REPONAME/$CIRCLE_PROJECT_REPONAME-integration.xml" -- -timeout 20m:
+      timeout: 10
```
dansanduleac: timeouts don't seem to match, why don't we just use circle's timeout?

init/cli/start.go
```diff
@@ -15,48 +15,90 @@
...
+	Name: "start",
+	Usage: `
+Ensures the service defined by the static and custom configurations at service/bin/launcher-static.yml and
+var/conf/launcher-custom.yml is running and its outputs are redirecting to var/log/startup.log and other
+var/log/${SUB_PROCESS}-startup.log files. If successful, exits 0, otherwise exits 1 and writes an error message to
+stderr and var/log/startup.log.`,
+	Action: executeWithContext(start, truncOutputFileFlag),
+}
+
+func start(ctx cli.Context) (rErr error) {
```
dansanduleac: only name the error if you're using it

init/cli/lib.go
```diff
@@ -0,0 +1,118 @@
...
+	"gopkg.in/yaml.v2"
+
+	"github.com/palantir/go-java-launcher/launchlib"
+)
+
+const (
+	outputFileFlag       = os.O_CREATE | os.O_WRONLY
+	truncOutputFileFlag  = outputFileFlag | os.O_TRUNC
+	appendOutputFileFlag = outputFileFlag | os.O_APPEND
+	outputFileMode       = 0666
```
dansanduleac: This should probably be 644, no reason allowing everyone write access to the output file

init/cli/lib.go
```diff
@@ -0,0 +1,118 @@
...
+		if _, ok := runningProcs[name]; !ok {
+			notRunningCmds[name] = cmd
+		}
+	}
+	return &serviceStatus{notRunningCmds, writtenPids, runningProcs}, nil
+}
+
+func getPidfileInfo() (servicePids, map[string]*os.Process, error) {
+	pidfileBytes, err := ioutil.ReadFile(pidfile)
+	if err != nil && !os.IsNotExist(err) {
```
dansanduleac: double negative - can we use `os.IsExist(err)`?

dansanduleac: and pull that out in a `pidfileExists` variable so that the else condition below reads `!pidfileExists` rather than calling the same / similar os function twice

dansanduleac (changes requested):
Main concern I have is that we're running the goinit executable in the integration test, which means we can't mock time and have to wait 240 seconds in the test. There's gotta be a better way of testing this, so here's my proposal:

Since the whole cli is nicely written such that you only have to call os.Exit at the very end, we should
1) mock the stop clock using an abstraction kind of like the one described [here](https://medium.com/agrea-technogies/mocking-time-with-go-a89e66553e79) but enriched to support ticker and Timer.stop
2) integration test should call `cli.App().Run(...)` in a goroutine, then manipulate the fake clock (e.g. elapse 239 sec on the fake clock, assert goroutine didn't finish, elapse 1 or 2 more seconds, assert it did) and make assertions on the exitcode yielded from the goroutine's channel

circle.yml
gregakinman: this will be moot after the tests stop running for forever, but i needed to override both circle's timeout and `go test`'s timeout. the circle timeout wasn't being respected by circle, so as a hack the stop tests print output after every subtest...

gregakinman:
Synced on Slack but I think the hard part here is how do we "assert goroutine didn't finish" and "assert it did"

gregakinman:
Pretty stumped on these test flakes. They work on my local machine 100% of the time but are quite flaky on Circle. @dansanduleac the flakes seem to be happening when we check the channel before the timeout has passed. We could go back to what we were doing before (before we mocked the clock) where we just check after the timeout has passed that the exit code + stderr are correct instead of doing both that and also ensuring it’s still running before the timeout has passed. That would mean not testing the latter, so it just depends on how much time we want to put into trying to track down the flakes.

Also, orthogonal to this, we could mock the timeout instead of the clock, and eliminate the complexity of the clockwork stuff while still achieving the test runtime we want.

dansanduleac:
There were a couple of race conditions here:
* Starting the subprocess in the function but then waiting for it in a goroutine (not 100% sure this is a race but looks like one) - solved by starting it in the goroutine
* Not waiting for child to trap signal 15. This became apparent when I logged the WaitStatus of our processes, and when it worked locally it would still say e.g. of two unstoppable processes that one was `terminated` and the other was `killed`, whereas both should have been killed.
  Solved this by opening a new fd 3, and waiting until the process closes it before we can consider it 'ready' and give control back to the test

dansanduleac:
Nice thing about this now is we can even assert that the process was killed by SIGTERM / SIGKILL in the case of a stoppable, and unstoppable process, respectively.

dansanduleac (approved):

iamdanfox:
:+1: glad we've got this in, but in future we need to break features like these into smaller incremental PRs so that there's less cross-atlantic back and forth... possibly co-locating??

PR #59, opened 2018-08-13
Correct logging functionality

cbrockington:
* Consume ServiceLoggers such that go-java-launcher can retain it's expected
   behavior of always logging to standard out
 * Adds clarity to go-init intentionally only truncating log files once, and
   also correctly truncates all processes being restarted, not just the
   primary

In creating an ete test in the other PR I realised that https://github.com/palantir/go-java-launcher/pull/49 previously broke the expected functionality of the go-java-launcher.

There is some discussion about whether the changes of logging sub-processes to separate files for the launcher are desired, but given it changes the expectation the discussion should at least be had.  I think the fact it wasn't had is likely symptomatic of the previous design being confusion which I hope to also resolve here.


init/cli/cli.go
```diff
@@ -35,33 +34,36 @@ func App() *cli.App {
...
 	return func(ctx cli.Context) (rErr error) {
 		// Fall back to default stdout if error opening log file
-		outputParentDir := filepath.Dir(launchlib.PrimaryOutputFile)
-		if err := os.MkdirAll(outputParentDir, 0755); err != nil {
+		if err := os.MkdirAll(logDir, 0755); err != nil {
 			log.Printf("Encountered error during MkdirAll the output dir '%s': %v, falling back to stdout",
-				outputParentDir, err)
-			return action(ctx)
+				logDir, err)
+			return action(ctx, launchlib.NewSimpleWriterLogger(os.Stdout))
```
cbrockington: Fail if can't make.

iamdanfox (approved):
Looks good, just want that fail-fast stuff and a readme sentence!

gregakinman:
So to clarify, the 2 things broken were:

- `go-java-launcher` logged directly to the output file instead of to `stdout`
- `go-init` was only truncating the primary process's log file

?

cbrockington:
Yep.  There was an offline discussion about whether that was were we wanted `go-java-launcher` to in fact log which was driven by the fact it is still directly called in docker environments, and in those environments we do want to log to stdout.

PR #63, opened 2018-10-10
Print the status to stdout or stderr.

cbrockington:
In adherence with http://refspecs.linuxbase.org/LSB_3.1.1/LSB-Core-generic/LSB-Core-generic/iniscrptact.html
referenced in the README, status should "print the current status of the service".
Previously go-init will log status to a file however this is not the conventional
meaning of "print".

### DIFF
--- a/init/cli/status.go
+++ b/init/cli/status.go
@@ -15,6 +15,9 @@
 package cli
 
 import (
+	"fmt"
+	"os"
+
 	"github.com/palantir/pkg/cli"
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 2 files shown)


init/cli/status.go
```diff
@@ -35,27 +38,90 @@ If exit code is nonzero, writes an error message to stderr and var/log/startup.l
...
+
+	fmt.Println(matched.Description)
 	return nil
 }
+
+type ServiceState struct {
+	Description string
+	Applicable  func(serviceStatus *serviceStatus, err error) bool
+	ExitStatus  func(serviceStatus *serviceStatus, err error) (int, error)
+}
```
dansanduleac, line 119: Is this coupling of data + methods into a struct idiomatic go?

I'd prefer to have this be an interface instead.

cbrockington: I don't want to speak for the go community but I dislike the interface approach since that forces the creation of separate four separate types for each `ServiceState`, all of them being stateless singletons.

cbrockington: As a slightly less hand-wavey comment, I think that while there is data and a method, the methods are conceptually related to the data, but none of their logic is.  In some ways the point is that these functions are in fact functions and strictly not methods, but it's useful to have a construct which ties them too each other, similar to a named tuple, since a `State` involves a particular description, definition of whether its applicable, and generation of exit code.

Might be easier to talk over in person if that isn't clear.

dansanduleac (approved):

PR #73, opened 2019-01-18
[improvement] Separate pids.yaml into separate <name>.pid files representing each subcommand

cbrockington:
Gives consistency with the format before the addition of subprocesses, and
provides a much simpler way for external processes to access pid information
since they no longer have to parse the yaml.

### DIFF
--- a/launchlib/config.go
+++ b/launchlib/config.go
@@ -29,7 +29,7 @@ import (
 )
 
 var (
-	subProcessNamePattern = regexp.MustCompile("^[a-z-]+$")
+	processNamePattern = regexp.MustCompile("^[a-z-]+$")
 )
 
 type VersionedConfig struct {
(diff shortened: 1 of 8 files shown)


iamdanfox:
@cbrockington I thought the layout of the `var/run` directory was considered an implementation detail of go-java-launcher... Are you making this PR because someone is going to start relying on finding a file called `var/run/my-service.pid` ??  If so, please could we also commit to this on the README

launchlib/config.go
```diff
@@ -29,7 +29,7 @@ import (
 )
 
 var (
-	subProcessNamePattern = regexp.MustCompile("^[a-z-]+$")
+	processNamePattern = regexp.MustCompile("^[a-zA-Z_-]+$")
```
iamdanfox: Why relax the regex? TODO: confirm with sls-packaging's regex ??

launchlib/config_test.go
```diff
@@ -125,7 +127,7 @@ subProcesses:
 				VersionedConfig: VersionedConfig{
 					Version: 1,
 				},
-				ServiceName: "",
```
iamdanfox, line 128: Confirmed that sls-packaging actually does fill in sensible values here (not hard-coded to primary, but the actual service name)

launchlib/config.go
cbrockington: It looks like sls-packaging doesn't have a regex.  Happy to add this there as well, though that would technically be an sls-packaging break.

iamdanfox: OK, let's keep it at this original regex for now! We can always follow-up with an sls-packaging PR later.

iamdanfox (approved):

PR #76, opened 2019-02-13
[improvement] javaHome accepts $JAVA_11_HOME style values

iamdanfox:
## Before this PR

We want to allow people to use a java installation whose location is supplied by another environment variable, e.g. $JAVA_11_HOME but this indirection is currently impossible. When choosing your desired java home, you had to specify either:

- a literal path on disk
- an empty string, which results in the value of $JAVA_HOME

## After this PR

You can now specify the following and everything should Just Work:

```yaml
javaHome: $JAVA_11_HOME
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,8 @@ configType: java
 configVersion: 1
 # REQUIRED - The main class to be run
 mainClass: my.package.Main
-# OPTIONAL - Path to the JRE, defaults to the JAVA_HOME environment variable if unset
-javaHome: javaHome
+# OPTIONAL - Path to the JRE or environment variable name (e.g. $JAVA_11_HOME). Defaults to the JAVA_HOME environment variable if unset
+javaHome: /opt/palantir/jdk8/Contents/Home
 # REQUIRED - The classpath entries; the final classpath is the ':'-concatenated list in the given order
 classpath:
   - ./foo.jar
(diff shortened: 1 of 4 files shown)


launchlib/launcher_test.go
```diff
@@ -38,6 +38,17 @@ func TestGetJavaHome(t *testing.T) {
...
 }
 
+func TestGetJavaHome_allowsReadingOtherEnvVar(t *testing.T) {
+	original := os.Getenv("SOME_VAR")
+	defer func() { require.NoError(t, os.Setenv("SOME_VAR", original)) }()
+
+	require.NoError(t, os.Setenv("SOME_VAR", "foo"))
+
+	javaHome, javaHomeErr := getJavaHome("$SOME_VAR")
+	assert.NoError(t, javaHomeErr, "getJavaHome correctly returns nil")
```
dansanduleac: message is confusing

dansanduleac (approved):

README.md
```diff
@@ -17,8 +17,8 @@ configType: java
 configVersion: 1
 # REQUIRED - The main class to be run
 mainClass: my.package.Main
-# OPTIONAL - Path to the JRE, defaults to the JAVA_HOME environment variable if unset
-javaHome: javaHome
+# OPTIONAL - Path to the JRE or environment variable name (e.g. $JAVA_11_HOME). Defaults to the JAVA_HOME environment variable if unset
```
markelliot, line 20: $JAVA_11_HOME or "JAVA_11_HOME"? We need to make sure that environment variables passed here are not evaluated (ie we need to disallow `$(rm -rf /)`)

iamdanfox: Yep don't worry, the dollar will only ever appear in a `launcher-static.yml` on disk and then immediately expanded by go, it never gets put in a shell!

markelliot: reading the Go code, it seems like you might eval the env var though? Could we have a test to check that `$(<stuff>)` definitely doesn't execute?

PR #84, opened 2019-04-09
[fix] Trim whitespace from pidfile content

bmoylan:
## Before this PR
We saw the below error in production:
```
failed to determine service status: failed to determine running processes: pid file did not contain an integer: strconv.Atoi: parsing "12773\n": invalid syntax
```

From reading the `Start` code I'm not sure how the newline got in there, but this should protect us from it in the future.

## After this PR
We will ignore spurious whitespace in the pidfile

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/go-java-launcher/84)

### DIFF
--- a/init/cli/lib.go
+++ b/init/cli/lib.go
@@ -21,6 +21,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"syscall"
 
 	"github.com/palantir/pkg/cli"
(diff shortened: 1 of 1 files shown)


iamdanfox (approved):

iamdanfox:
Looking at other calls to Atoi, it seems they're pulling strings out of command line args which are provided by go-java-launcher itself, so it's probably fine to leave them as-is.

```
launcher/main/main.go
42:	if monitor.PrimaryPID, err = strconv.Atoi(primaryPID); err != nil {
47:		pid, err := strconv.Atoi(pidStr)
```

PR #92, opened 2019-06-14
[fix] SIGKILL after 2 minutes, not 4

iamdanfox:
## Before this PR

Currently, a java process which is stuck for some reason will never receive SIGKILL because our deployment infra only allows shutdowns to take 2 minutes. If deployment binary then SIGKILL's go-java-launcher, which was still waiting for it's 4 minute timeout to elapse, then underlying java process never actually gets killed properly.

## After this PR

We can't catch the incoming SIGKILL for the parent deployment process, so instead our timeout is slightly shorter (110 seconds), so that we can fire off a SIGKILL to the child java process rather than leaving it as a zombie.

## Possible downsides

- I don't think any WC services can be relying on the 4 minute timeout because WC itself already has a 2 minute timeout!!

### DIFF
--- a/init/cli/stop.go
+++ b/init/cli/stop.go
@@ -38,7 +38,7 @@ var stopCliCommand = cli.Command{
 	Usage: `
 Ensures the service defined by the static and custom configurations are service/bin/launcher-static.yml and
 var/conf/launcher-custom.yml is not running. If successful, exits 0, otherwise exits 1 and writes an error message to
-stderr and var/log/startup.log. Waits for at least 120 seconds for any processes to stop before sending a SIGKILL.`,
+stderr and var/log/startup.log. Waits for at least 110 seconds for any processes to stop before sending a SIGKILL.`,
 	Action: executeWithLoggers(stop, NewAlwaysAppending()),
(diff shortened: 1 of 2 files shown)


init/cli/stop.go
```diff
@@ -95,7 +95,7 @@ func stopService(ctx cli.Context, procs map[string]*os.Process) error {
 }
 
 func waitForServiceToStop(ctx cli.Context, procs map[string]*os.Process) error {
-	const numSecondsToWait = 240
+	const numSecondsToWait = 120
```
CRogers: It might be safer to give a little more time to get in there before the deployment infra does? Like 115 seconds? Otherwise if the deployment infra is a little faster with the sigkill than the sigterm we have the same problem?

CRogers (approved):

iamdanfox:
So I think my one concern before actually merging this is I can't figure out why our deployment platform would actually _interrupt_ go java launcher early, as it does seem to have a 250second timeout.

EDIT also, it seems it just calls the "stop" command here, so this is likely not the real solution.

PR #94, opened 2019-06-18
Make develop green again

iamdanfox:
The most recent commit on develop was 2 months ago, which was green (https://circleci.com/gh/palantir/go-java-launcher/651).

Unfortunately, I think CircleCI changed something upstream, because we've seen similar problems with unit-tests that involved subprocesses in docker containers in sls-packaging (https://github.com/palantir/sls-packaging/pull/508).


ferozco:
You're probably right, its the exact same problem as what hit sls-packaging

CRogers:
Probably would want to use something like https://github.com/Yelp/dumb-init if we really want to fix this.

CRogers (approved):
Approving for now as this test is already horribly broken

PR #109, opened 2019-09-06
[improvement] Change default behavior of 'javaHome' option

atrocities:
This option previously meant 'Use $JAVA_HOME' if this option is not set
at all. It now means 'use $JAVA_HOME' if this option is not set OR if
the environment variable it's set to is not defined at runtime.

This is useful in environments where a default JDK is preferred, but
init should not fail should this JDK be absent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,8 @@ configType: java
 configVersion: 1
 # REQUIRED - The main class to be run
 mainClass: my.package.Main
-# OPTIONAL - Path to the JRE or environment variable name (e.g. $JAVA_11_HOME). Defaults to the JAVA_HOME environment variable if unset
+# OPTIONAL - Path to the JRE or environment variable name (e.g. $JAVA_11_HOME). Defaults to the JAVA_HOME environment variable if unset,
+#  or if the specified environment variable is not set at runtime.
 javaHome: /opt/palantir/jdk8/Contents/Home
 # REQUIRED - The classpath entries; the final classpath is the ':'-concatenated list in the given order
 classpath:
--- a/changelog/@unreleased/pr-109.yml
+++ b/changelog/@unreleased/pr-109.yml
@@ -1,0 +1,7 @@
+type: improvement
+improvement:
+  description: The javaHome option, when set to a variable, now
+    checks to see if the variable is set at runtime. If it is
+    not set, it falls back on $JAVA_HOME.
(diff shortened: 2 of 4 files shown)


iamdanfox:
I'm not super keen on this, because now when an operators looks at their product/launcher-static.yml, they can no longer say with certainty whether a service is actually running with Java11 or not.

:-1:

atrocities:
@iamdanfox - understood. What is your preferred way of implementing an "if not set then use default" mechanism?

chrismale:
@iamdanfox we are in a difficult position right now of needing to run services which straddle environment versions.  Some environments have an environmental variable pointing to a JDK we want to use, some older environments do not.  We want to change our products to use the new variable but have a fall back to the older JAVA_HOME for the older environments.

carterkozak:
Rather than always falling back to `JAVA_HOME`, what if we allowed a preference ordered list of acceptable locations/envvars?

chrismale:
+1 to @carterkozak's suggestion.  That provides some clarity for developers about what Java environment maybe used for their services and is a no-op for products which do not want to use a fallback list.

iamdanfox:
Ok so @jlong8 explained over slack that the motivation for this is actually unrelated to Java11, it's to prefer a zulu enterprise java8 installation (from `$JAVA_ZULU_8_HOME`) if it's available, and fall back to `$JAVA_HOME` otherwise.  I still don't really want anyone in Foundry to start relying on this, but if this is particularly time sensitive, then I'd be OK with merging this hack if we're OK with reverting it in a month or two?

carterkozak:
I imagine this will come up every year or so for products which cannot depend on frequent Skylab upgrades, so a more general solution may be preferable.

chrismale:
Hey @carterkozak are you good with this doing this initial change to unblock us today? Agree on the long term needs.

jlong8:
I chatted offline with @chrismale, @iamdanfox, and @atrocities, and I think we're good to merge, with the explicit commitment that we will revert this in 2 months.

@iamdanfox could you approve?

iamdanfox (approved):

iamdanfox:
If you add a proper file to the changelog/@unreleased dir then we could autorelease.

atrocities:
Hmmm... is something up with the changelog check?

PR #125, opened 2020-01-28
Add build outputs for ARM

markelliot:
We have some new needs to deploy services on ARM, so we should begin packaging these artifacts.

## Before this PR
Builds for Mac and Linux on amd64.

## After this PR
==COMMIT_MSG==
Build outputs include arm64
==COMMIT_MSG==

## Possible downsides?
Additional storage cost of built artifacts.


nmiyake (approved):
Just chiming in to say that the change looks functionally correct (should do what is stated as desired) from my end

ferozco:
👍 defer to @nmiyake here

Issue #128, opened 2020-04-13
Add ability to launch into cgroups


a10y:
Feature Request: add ability to launch into cgroups.

Internally we have a number of services that call into native libraries from Java. When either the FFI code or the actual native code has bugs, this can occasionally lead to memory leaks which don't surface at the JVM level as OutOfMemoryErrors, but instead continue to cause process RSS to rise until it eventually consumes all physical memory in the system.

An easy way to prevent this from happening is to place the process into a cgroup (which is what docker/other container runtimes do behind the scenes) and set fine-grained memory limits on a process.

Adding a process into a cgroup is really easy: you just write a PID into a pseudo-file, e.g.

```
$ top &
$ echo $! > /sys/fs/cgroup/memory/my-group/tasks
```

I'd like to add the ability to allow go-java-launcher to launch a process into a pre-created cgroup, which can be configured via the `launcher-custom.yml` file.

a10y:
Would likely only support cgroupsv1 as a first pass, and likely only memory controllers. pseudo-conjure syntax:

```yaml
CgroupConfig:
  enum:
    - CgroupV1Config
CgroupV1Config:
  fields:
    memoryGroup: string
```

a10y:
Merged in #129

PR #129, opened 2020-04-13
Enable cgroupsv1 config

a10y:
Implementation of #128

### DIFF
--- a/launchlib/launcher.go
+++ b/launchlib/launcher.go
@@ -45,7 +45,7 @@ func CompileCmdsFromConfig(
 		SubProcesses: make(map[string]*exec.Cmd),
 	}
 
-	serviceCmds.Primary, err = compileCmdFromConfig(&staticConfig.StaticLauncherConfig, &customConfig.CustomLauncherConfig, loggers.PrimaryLogger)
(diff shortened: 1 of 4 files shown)


a10y:
Tested manually, appears to work as expected. Test plan:

* Create control group: `cgcreate -t user -a user memory:test1`
* Set an 800MB limit for the group: `echo 800000000 > /sys/fs/cgroup/memory/test1/memory.limit_in_bytes`
* Add the launcher-config overrides for the service we want launched:
```
launcher-custom:
  cgroupsV1:
    memory: test1
```
* Restart the service using the go-init binary
* Check on `/sys/fs/cgroup/memory/test1/memory.usage_in_bytes`

The test service is also running on jdk13, which respects the cgroup max memory, so even though it has `heap-size: 1g` it's only occupying 799MB

launchlib/launcher.go
```diff
@@ -115,6 +116,18 @@ func compileCmdFromConfig(staticConfig *StaticLauncherConfig, customConfig *Cust
 	}
 
 	args = append(args, staticConfig.Args...)
+	if len(*cgroupsV1) > 0 {
+		var cgexecArgs []string
+		executable = "/bin/cgexec"
```
a10y, line 121: believe this should be relatively portable, at least between rhel/ubuntu

carterkozak:
:+1: looks reasonable to me.

ferozco:
Sure lets give it a shot 👍

PR #130, opened 2020-06-09
fix process running check

rzpt:
If the pid we are checking is actually a currently running tid, then
kill will return success. We have to filter against actually running
processes. Iterating over entries in /proc will filter to actual pids
and not return any tids.

## Before this PR

If a tid matches the pid we are checking against, then kill will return non nil, and so go-init will incorrectly flag the process as running.

## After this PR
==COMMIT_MSG==
Fix false positive process detection when an unrelated thread's id matches the old process id.
==COMMIT_MSG==

## Possible downsides?

I am not sure how to handle the darwin case, or if the bug even repros on darwin. In this PR, the darwin check is the same as the linux check, but it could possibly be omitted. I also added an extra error return value where there previously was not one.

There is also this bug which this PR does not handle: We also do not check for a different process that has taken over the old process's pid.

### DIFF
--- a/init/cli/lib.go
+++ b/init/cli/lib.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 	"syscall"
 
+	ps "github.com/mitchellh/go-ps"
 	"github.com/palantir/pkg/cli"
 	"github.com/pkg/errors"
 
@@ -108,7 +109,11 @@ func getCmdProcess(name string) (*int, *os.Process, error) {
 		return nil, nil, errors.Wrap(err, "pid file did not contain an integer")
 	}
 
-	if running, proc := isPidRunning(pid); running {
+	running, proc, err := isPidRunning(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+	if running {
 		return &pid, proc, nil
 	}
 	return &pid, nil, nil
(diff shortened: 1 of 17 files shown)


init/cli/lib.go
```diff
@@ -146,16 +151,37 @@ func getConfiguredCommands(ctx cli.Context, loggers launchlib.ServiceLoggers) (m
...
 	// This is the way to check if a process exists: https://linux.die.net/man/2/kill.
-	return proc.Signal(syscall.Signal(0)) == nil
+	// On linux, this may respond true if there is a thread running with the same id.
+	running := proc.Signal(syscall.Signal(0)) == nil
+	if !running {
+		return false, nil
+	}
+
+	// On linux, iterating over processes will only return running processes. Unfortunately,
+	// getting exactly the one pid will also return a thread of the same id.
```
dansanduleac, line 176: > Unfortunately, getting exactly the one pid will also return a thread of the same id.

Is there actually a difference between calling `ps.Processes()` then manually looking for a pid, vs calling `findProcess(pid)`?

From the [process_unix.go](https://github.com/mitchellh/go-ps/blob/master/process_unix.go#L36-L90) of this library it seems to me that both will end up doing the same thing, which is looking in `/proc/` for a directory(ies) matching a process number, and then wrap that with `newUnixProcess`.

rzpt: Yes. I tested it manually, they are different.

From man proc:
```
The /proc/[tid] subdirectories are not visible when iterating through /proc with getdents(2) (and thus are not visible when one uses ls(1) to view the contents of /proc).
```

dansanduleac (approved):
Alright then, let's give it a go. Tried to add changelog, but the bot won't run because you've made the PR from a fork.

Can you write the changelog please?

PR #135, opened 2020-08-26
Upgrade godel and fix lints

bmoylan:
Includes #134 and #122

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/go-java-launcher/135)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,125 +2,52 @@
 # To request a modification to the general template, file an issue on Excavator.
 # To manually manage the CircleCI configuration for this project, remove the .circleci/template.sh file.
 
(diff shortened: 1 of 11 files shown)


bmoylan:
cc @ferozco

ferozco:
Thanks @bmoylan! Looks like the circle-template assumes go-modules. Should we also do that migration here?

bmoylan:
We should do it, but probably in a different PR. The failing circle tasks are just about caching and they don't fail the builds. I've pushed `bm/modules` but it has some weird failure I don't see locally so have to look into that

ferozco:
👍 cool happy to move forward with this. Thanks again!

PR #141, opened 2021-02-10
rotate startup logs

rzpt:
## Before this PR

When a service fails to start, it is automatically started again. A new startup log is written, overwriting the existing one, which makes it impossible to determine why a service failed to start if it crashes early enough in the startup process.

## After this PR

==COMMIT_MSG==
Startup logs are rotated. 6 old logs are retained.
==COMMIT_MSG==

## Possible downsides?

Extra disk space usage.

### DIFF
--- a/init/cli/logging.go
+++ b/init/cli/logging.go
@@ -97,11 +97,11 @@ func (f *FileLoggers) OpenFile(path string) (*os.File, error) {
 
 func rotate(path string) {
 	limit := 5
-	os.Remove(path + "." + strconv.Itoa(limit))
+	_ = os.Remove(path + "." + strconv.Itoa(limit))
 	for i := limit; i > 0; i-- {
-		os.Rename(path+"."+strconv.Itoa(i-1), path+"."+strconv.Itoa(i))
+		_ = os.Rename(path+"."+strconv.Itoa(i-1), path+"."+strconv.Itoa(i))
 	}
-	os.Rename(path, path+".0")
+	_ = os.Rename(path, path+".0")
 }
 
 var devNull = launchlib.NoopClosingWriter{Writer: ioutil.Discard}
--- a/init/cli/logging.go
+++ b/init/cli/logging.go
@@ -25,6 +25,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+const limit = 5
+
 type FileFlags interface {
 	Get(path string) int
 }
@@ -60,14 +62,14 @@ func NewAlwaysAppending() FileFlags {
 type FileLoggers struct {
 	flags   FileFlags
 	mode    os.FileMode
-	rotated map[string]bool
+	rotated map[string]struct{}
 }
 
 func NewFileLoggers(flags FileFlags, mode os.FileMode) *FileLoggers {
 	return &FileLoggers{
 		flags:   flags,
 		mode:    mode,
-		rotated: make(map[string]bool),
+		rotated: make(map[string]struct{}),
 	}
 }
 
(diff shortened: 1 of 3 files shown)


carterkozak:
I'm in favor of this change, though I'd want someone more familiar with golang than myself to give it a review.

init/cli/logging.go
```diff
@@ -72,13 +82,28 @@ func (f *FileLoggers) SubProcessLogger(name string) launchlib.CreateLogger {
...
 }
 
+func rotate(path string) {
+	limit := 5
+	_ = os.Remove(path + "." + strconv.Itoa(limit))
+	for i := limit; i > 0; i-- {
+		_ = os.Rename(path+"."+strconv.Itoa(i-1), path+"."+strconv.Itoa(i))
+	}
+	_ = os.Rename(path, path+".0")
+}
```
carterkozak, line 106: Alternatively we could rely on mtimes + always increment the index to avoid renaming N files per startup, but that makes the code more complex. Unsure if we should read and log errors if they exist, I suppose we'd have to check if files exist before attempting renames in that case.

init/cli/logging.go
```diff
@@ -72,13 +82,28 @@ func (f *FileLoggers) SubProcessLogger(name string) launchlib.CreateLogger {
...
 	return file, nil
 }
 
+func rotate(path string) {
+	limit := 5
+	_ = os.Remove(path + "." + strconv.Itoa(limit))
+	for i := limit; i > 0; i-- {
+		_ = os.Rename(path+"."+strconv.Itoa(i-1), path+"."+strconv.Itoa(i))
+	}
+	_ = os.Rename(path, path+".0")
```
carterkozak, line 105: Spacing looks odd here, isn't there an auto-formatter that's supposed to automatically check and fix these things?

rzpt: This was the autoformatter. I think when a function has multiple arguments, then the spaces are removed. You can see on line 100 above, the function takes one argument and there are spaces.

init/cli/logging.go
```diff
@@ -57,8 +58,17 @@ func NewAlwaysAppending() FileFlags {
 }
 
 type FileLoggers struct {
-	flags FileFlags
-	mode  os.FileMode
+	flags   FileFlags
+	mode    os.FileMode
+	rotated map[string]bool
```
tabboud: nit: It's generally best practice to implement sets using `map[string]struct{}`, where existence in the map is sufficient.

Also I assume this logger is not used concurrently?

init/cli/logging.go
```diff
@@ -72,13 +82,28 @@ func (f *FileLoggers) SubProcessLogger(name string) launchlib.CreateLogger {
...
+	}
 	file, err := os.OpenFile(path, f.flags.Get(path), f.mode)
 	if err != nil {
 		return file, errors.Wrapf(err, "could not open logging file '%s'", path)
 	}
 	return file, nil
 }
 
+func rotate(path string) {
+	limit := 5
```
tabboud: nit: explicitly declare this a const i.e. `const limit = 5`

init/cli/logging.go
```diff
@@ -72,13 +82,28 @@ func (f *FileLoggers) SubProcessLogger(name string) launchlib.CreateLogger {
 }
 
 func (f *FileLoggers) OpenFile(path string) (*os.File, error) {
+	if _, ok := f.rotated[path]; !ok {
```
tabboud, line 87: Is `OpenFile` called more than once for a single path? I am just curious if you actually have to track what has been rotated and instead just always rotate on a call to `OpenFile`

init/cli/logging.go
rzpt: Done.

init/cli/logging.go
rzpt: Yes. Without doing something like this, the startup file was being rotated multiple times per invocation of the launcher.

init/cli/logging.go
rzpt: Each returned logger is not used concurrently, but I am not sure if multiple loggers to the same path are used concurrently.

tabboud (commented):
👍 All good from me and will defer to @carterkozak for final approval

tpetracca (approved):
thanks for the extra set of eyes @tabboud; I'm happy to merge this as is then

PR #154, opened 2021-05-27
Container support for java

jkozlowski:
## Before this PR

Developers have to manually specify heap size when running in containers.

## After this PR

==COMMIT_MSG==
When running inside container, java heap is configured automatically to take 80% of available memory.
==COMMIT_MSG==


launchlib/launcher.go
```diff
@@ -265,3 +281,18 @@ func createReplacer() *strings.Replacer {
...
+		if !isHeapArg(arg) {
+			filtered = append(filtered, arg)
+		}
+	}
+	return filtered
+}
+
+func isHeapArg(arg string) bool {
+	var lowerCase = strings.ToLower(arg)
+	return strings.HasPrefix(lowerCase, "-xmx") || strings.HasPrefix(lowerCase, "-xms")
```
carterkozak: I think case is required to match `-Xmx` and `-Xms`

jkozlowski: Oh you mean nobody would do -xmx? fair, I wasn't sure if this stuff is case sensitive

launchlib/launcher.go
```diff
@@ -95,13 +95,24 @@ func compileCmdFromConfig(
 			staticConfig.JavaConfig.Classpath))
 		_, _ = fmt.Fprintf(logger, "Classpath: %s\n", classpath)
 
+		var jvmOpts []string
+
+		if isEnvVarSet("CONTAINER") && staticConfig.ContainerSupport {
+			_, _ = fmt.Fprintln(logger, "Container support enabled")
+			jvmOpts = append(jvmOpts, filterHeapArgs(staticConfig.JavaConfig.JvmOpts)...)
+			jvmOpts = append(jvmOpts, filterHeapArgs(customConfig.JvmOpts)...)
+			jvmOpts = append(jvmOpts, []string{"-XX:+UseContainerSupport", "-XX:InitialRAMPercentage=80.0", "-XX:MaxRAMPercentage=80.0"}...)
```
carterkozak: Brainstorm: I wonder if we should approach this a little bit differently? Could we replace all our `-Xms` and `-Xmx` configuration parameters with `-XX:MaxRAM=<value>`, then we only have to filter out `MaxRAM` in containers where we can trust the detected value.
If we're already requiring folks to set `containerSupport` in their static config, it might work nicely.

carterkozak: In that case we wouldn't need additional config, if CONTAINER is set, we strip `MaxRAM`, however current services aren't impacted by that, and behavior isn't modified.

jkozlowski: I haven't seen MaxRAM before: so you're suggesting parsing Xmx to figure out bytes to pass to MaxRAM, WHEN containerSupport isn't set? and when containerSupport is on, just filtering Xmx/Xms and letting the JVM do the right thing?

jkozlowski: I have read some random people on the internet suggesting MaxRAMPercentage is a safer bet: https://medium.com/adorsys/jvm-memory-settings-in-a-container-environment-64b0840e1d9e

iamdanfox:
Oh my days is this going to solve all my oomkilled k8s services from mis-tuned heaps??  (If necessary, we can probably volunteer some services as tribute to try out the new JVM opts before giving them to the whole fleet)

launchlib/launcher.go
jkozlowski: From what I'm reading, if you're only setting MaxRAM, it will use MaxRAMFraction to actually figure out the heap size, which by default is 4, which means you're giving yourself 25% of memory available in the container. So we'd need to adjust that anyway? I personally prefer to read percentages than 1/MaxRAMFraction to figure out what it's going to give to the process.

carterkozak: Idea is:

Excavate out a change to replace `-Xms` and `-Xmx` with `-XX:MaxRAM=<1.25*Xmx> -XX:MaxRAMPercentage=80 -XX:InitialRAMPercentage=80*(original_xms / original_xmx)` 

Then, all we have to do in the launcher is filter out `-XX:MaxRAM` when the `CONTAINER` envvar is present, we don't have to worry about xmx and xms at all.

carterkozak: Note the numbers are based on the assumption that heap should be 80% of total ram, that may be inaccurate in which case we'd need to update values accordingly.

jkozlowski:
@iamdanfox that's the dream, I was just about to start hunting for some services to test this out on. Need to finish up https://github.com/palantir/sls-packaging/pull/1147 first

launchlib/launcher.go
jkozlowski: That sounds like a lot more work to me, for not much gain? what's wrong with MaxRAMPercentage? (I am definitely assuming that it actually works, but to do that I will actually have to test it out, so it might be that your way is the only way)

jkozlowski: Oh, re-reading this a little bit: you are suggesting using MaxRAMPercentage, sorry. I mean I don't really mind, just feels like this is nice and hidden here, nobody has to worry about it, and it does the right thing. I'd have to have some filtering logic anyway, so if I have to filter out MaxRAM or xmx/xms what difference does it make? I guess I should make sure this is also filtering maxram...

jkozlowski: In general, I think the assumption we're all making is that in general we'll be running with pretty small heaps, so sacrificing 20% does not feel bad. But for something like 8GB heap, assuming that native memory grows linearly is maybe not an always bullet-proof assumption? but it's probably good enough to work in most cases.

I feel we should probably have a metric that allows us to see actual heap usage somehow

carterkozak: Filtering out `MaxRAM` is easier imo than replacing xms and xmx with another pair of parameters, I like the idea of keeping this project as simple as possible (remove exactly one parameter, if it exists). I like the idea of reusing as much configuration as possible in non-containerized environments as well.

> I think the assumption we're all making is that in general we'll be running with pretty small heaps, so sacrificing 20% does not feel bad. But for something like 8GB heap, assuming that native memory grows linearly is maybe not an always bullet-proof assumption? but it's probably good enough to work in most cases.

We need to worry about nonheap memory as well: thread overhead, direct memory (ByteBuffer.allocateDirect, defaults to a maximum equal to your heap), native jni overhead, metaspace, etc. Some fun detail here: https://shipilev.net/jvm/anatomy-quarks/12-native-memory-tracking/

> I feel we should probably have a metric that allows us to see actual heap usage somehow

Our memory pool metrics from tritium give us this data, but not in the easiest to consume formats (have to sum the pool sizes).

jkozlowski: Ok do my final thing is that: even if we excavate it out, you’d still have launcher-custom, which you cannot excavate?

carterkozak: Yep, if you explicitly set custom heap, that's still used. I think that's okay and more obvious.

jkozlowski: The more I think about it, the more I feel like we should have different distributions: classic tar, container and container with native image. And it shouldn’t be the go-Java-launcher that deals with this complexity, but it’s dealt with at build time.

jkozlowski: The gradle distribution plugin supports multiple distributions nicely: https://docs.gradle.org/current/userguide/distribution_plugin.html

jkozlowski: > I think that's okay and more obvious.

I think the problem with this is that ALL products have this overriden, because that's how you set heap right now. Most products just map launcherCustom.heap to the launcher-custom.yml file and call it a day. Virtually no products use "defaultJvmArgs" from sls-packaging to set their default heap sizes.

carterkozak: re multiple distribution types based on container/non-container: I think this runs into an issue with our config/manifest distribution, which is expected to match the configuraiton for a distribution with the same coordinates. This makes it difficult to support divergent configuration schemes based on the deployment model. I don't know the specifics though, and I may be incorrect about how it works.

jkozlowski: After much deliberating, we have settled on this being the best solution to the problem, where go-java-launcher is reponsible for unsetting -Xmx and -Xms, and setting MaxRamPercentage/InitialRAMPercentage if both not set.

launchlib/launcher.go
```diff
@@ -265,3 +282,37 @@ func createReplacer() *strings.Replacer {
...
+		if isMaxRAMPercentage(arg) {
+			hasMaxRAMPercentage = true
+		} else if isInitialRAMPercentage(arg) {
+			hasInitialRAMPercentage = true
+		}
+	}
+
+	if !hasInitialRAMPercentage && !hasMaxRAMPercentage {
+		filtered = append(filtered, "-XX:InitialRAMPercentage=80.0")
+		filtered = append(filtered, "-XX:MaxRAMPercentage=80.0")
```
iamdanfox: @jkozlowski I hadn't actually realized this PR involves making go-java-launcher introspect the command line args defined in launcher-custom.yml and replace out some of them.  I think this is honestly gonna confuse the crap out of people, because up until now the contract of go-java-launcher has been that it effectively just invokes the `java` binary with all the static & custom args laid out afterwards.

What do you think about having an extra section in launcherCustom or launcher-custom.container.yml or something that you'd expect folks to populate, and then go-java-lancher would just switch between them based on the $CONTAINER var?

jkozlowski: Discussed in person, this approach is still safest and cleanest out of all the proposed solutions.

integration_test/go_java_launcher_integration_test.go
```diff
@@ -38,11 +38,35 @@ func TestMainMethod(t *testing.T) {
 	require.NoError(t, err, "failed: %s", output)
 
 	// part of expected output from launcher
-	assert.Regexp(t, `Argument list to executable binary: \[.+/bin/java -Xmx4M -Xmx1g -classpath .+/github.com/palantir/go-java-launcher/integration_test/testdata Main arg1\]`, output)
+	assert.Regexp(t, `Argument list to executable binary: \[.+/bin/java -Xmx4M -Xmx1g -classpath .+/integration_test/testdata Main arg1\]`, output)
```
jkozlowski: When running this test locally, the path is rendered as per where my it is cloned on my, so I remove the unnecessary bits.

carterkozak (approved):
The design and logic look good to me, but I defer to folks more familiar with golang for review on the implementation details.

integration_test/go_java_launcher_integration_test.go
```diff
@@ -38,11 +38,35 @@ func TestMainMethod(t *testing.T) {
...
+	testContainerSupportEnabled(t, "testdata/launcher-custom-initial-ram-percentage-override.yml", "-XX\\:InitialRAMPercentage=79.9")
+}
+
+func TestMainMethodJavaContainerSupportLauncherCustomMaxRamPercentageOverride(t *testing.T) {
+	testContainerSupportEnabled(t, "testdata/launcher-custom-max-ram-percentage-override.yml", "-XX\\:MaxRAMPercentage=79.9")
+}
+
+func TestMainMethodJavaContainerSupportLauncherCustomInitialAndMaxRamPercentageOverride(t *testing.T) {
+	testContainerSupportEnabled(t, "testdata/launcher-custom-initial-and-max-ram-percentage-override.yml", "-XX\\:InitialRAMPercentage=79.9 -XX\\:MaxRAMPercentage=80.9")
+}
```
tabboud: note: This is totally fine to keep all tests as their own function, however one thing you will typically see in scenarios like this would be to use a [table-driven test](https://github.com/golang/go/wiki/TableDrivenTests), which looks something like this.
```go
func TestMainMethodJavaContainerSupport(t *testing.T) {
	for _, tc := range []struct {
		name            string
		launcherCustom  string
		expectedJVMArgs string
	}{
		{
			name:            "sets defaults",
			launcherCustom:  "testdata/launcher-custom.yml",
			expectedJVMArgs: "-XX\\:InitialRAMPercentage=80.0 -XX\\:MaxRAMPercentage=80.0",
		},
		// ... Other test cases
	} {
		t.Run(tc.name, func(t *testing.T) {
			testContainerSupportEnabled(t, tc.launcherCustom, tc.expectedJVMArgs)
		})
	}
}

```
No need to change this, but just wanted to give you some extra context.

integration_test/go_java_launcher_integration_test.go
```diff
@@ -114,8 +143,21 @@ func TestSubProcessesParsedMonitorSignals(t *testing.T) {
...
 }
 
-func runMainWithArgs(t *testing.T, staticConfigFile, customConfigFile string) (string, error) {
-	output, err := mainWithArgs(t, staticConfigFile, customConfigFile).CombinedOutput()
+func runMainWithArgs(t *testing.T, staticConfigFile, customConfigFile string, env ...string) (string, error) {
+	jdkDir := "jdk"
+	javaHome, err := filepath.Abs(jdkDir)
+	if err != nil {
+		log.Fatalf("Failed to calculate absolute path of '%s': %v\n", jdkDir, err)
+	}
```
tabboud: nit: Given you have access to the `*testing.T` you can alternatively use `require.NoErrorf(t, err, "Failed to calculate ...", jdkDir)` instead of fatal logging.

integration_test/go_java_launcher_integration_test.go
```diff
@@ -114,8 +143,21 @@ func TestSubProcessesParsedMonitorSignals(t *testing.T) {
...
+
+	// Override existing environment when running subprocess.
+	var customEnv []string
+	customEnv = append(customEnv, "JAVA_HOME="+javaHome)
+	customEnv = append(customEnv, env...)
```
tabboud: note: you can do this in 1 line: `customEnv := append([]string{"JAVA_HOME="+javaHome}, env...)`

launchlib/launcher.go
```diff
@@ -95,13 +95,18 @@ func compileCmdFromConfig(
 			staticConfig.JavaConfig.Classpath))
 		_, _ = fmt.Fprintf(logger, "Classpath: %s\n", classpath)
 
+		var combinedJvmOpts []string
+		combinedJvmOpts = append(combinedJvmOpts, staticConfig.JavaConfig.JvmOpts...)
+		combinedJvmOpts = append(combinedJvmOpts, customConfig.JvmOpts...)
```
tabboud, line 100: note: This can be 1 line as well: 
```go
combinedJvmOpts := append(staticConfig.JavaConfig.JvmOpts, customConfig.JvmOpts...)
```

tabboud (commented):
Overall looks good to me. I left a few notes on Go style/best-practices, but none are blocking and was mostly informational

integration_test/go_java_launcher_integration_test.go
jkozlowski: ah, happy to change over to this style if it's more idiomatic?

launchlib/launcher.go
jkozlowski: Does this not modify staticConfig.JavaConfig.JvmOpts backing array/slice?

integration_test/go_java_launcher_integration_test.go
tabboud: It's generally more idiomatic, but again both ways are fine

launchlib/launcher.go
tabboud: Technically it can modify the backing array, but given both are slices it should be fine. It all depends on the size of the original slice `jvmOpts`. Let's just keep it the way you have it written to avoid any ambiguity in the backing array.

jkozlowski: yeah, I think it's rather not modify the originals, keep it nice and read-only.

integration_test/go_java_launcher_integration_test.go
jkozlowski: Done

integration_test/go_java_launcher_integration_test.go
jkozlowski: Done

integration_test/go_java_launcher_integration_test.go
jkozlowski: I am going to switch: I quite liket he ability to just define in-method structs so effortlessly.

jkozlowski: Should be done.

PR #156, opened 2021-06-04
[WIP] `containerJvmArgs` are used when running inside a container.

jkozlowski:
… container.

## Before this PR
Developers have no way to configure heap when running inside containers.

## After this PR

==COMMIT_MSG==
`containerJvmArgs` are used when running inside a container.
==COMMIT_MSG==


jkozlowski:
Closing, as decided against this approach.

Issue #313, opened 2023-03-10
Implement Improved Container Support


carterkozak:
Currently we support containers thusly:
https://github.com/palantir/go-java-launcher/blob/3e6f05325a8002960172b1bd0eb39aec0a5ba42f/README.md#java-heap-and-container-support

`-Xms` and `-Xmx` args are ignored, and replaced with `-XX:InitialRAMPercentage=75.0 -XX:MaxRAMPercentage=75.0` to set heap memory bounds dynamically based on the container itself. This is better than configuring the container and jvm heap separately, but has a number of problems:

1. In practice, our heap needs don't scale linearly. A container with 256m of memory will likely need a lower percentage for heap than a container with 32g of memory in order to make space for the jvms native memory requirements (which themselves scale, but not entirely linearly with heap).
2. `InitialRAMPercentage` and `MaxRAMPercentage` seem not to be entirely equivalent to `Xms` and `Xmx` in that unused heap may still be returned to the OS when both flags are set to the same value (Documentation isn't terribly helpful here, and this may need to be re-evaluated for correctness, but this is my recollection based on the original rollout)
2. CPU information is not read correctly by the JVM. It never was, but this became much worse when https://bugs.openjdk.org/browse/JDK-8281181 was shipped in the October Java CPU (Critical Patch Update). Where previously the JVM modified its active processor count based on detected cgroup information, it no longer does without a new (deprecated) flag: `-XX:+UseContainerCpuShares` https://github.com/palantir/sls-packaging/pull/1414
    * However, even with this feature enabled, services set to use 1000 millicores fall through a codepath which causes them to detect all cores on the hypervisor (due to ambiguity in whether or not cgroups are being leveraged, as cgroups are used more broadly than our use-case). See [cgroupV1Subsystem_linux.cpp](https://github.com/openjdk/jdk17u/blob/703a82a34381c04792ad5f5c6c677827c787fc1b/src/hotspot/os/linux/cgroupV1Subsystem_linux.cpp#L294-L295) or the similar case for cgroupsv2 in [CgroupV2Subsystem.java](https://github.com/openjdk/jdk17u/blob/5159ffd487d858f33eb50f0bb8f2bfb28fcce5ae/src/java.base/linux/classes/jdk/internal/platform/cgroupv2/CgroupV2Subsystem.java#L157-L158).

## Proposed Path Forward

We should teach go-java-launcher to detect the configurations that we use, where we have more specific knowledge about what we expect.

## Processors

[...]

PR #319, opened 2023-03-28
Add support for setting active processor argument

andybradshaw:
## Before this PR

Most of the context exists in https://github.com/palantir/go-java-launcher/issues/313, but to briefly summarize here, there are certain scenarios where using a very low CPU request can cause the JVM to make some assumptions around how many threads should be used for various tasks. We would like to move the CPU share detection logic into the launcher and pass that information to the JVM, so we have a little bit more control in those situations.

## After this PR

==COMMIT_MSG==
Add support for setting the `-XX:ActiveProcessorCount` argument driven by detecting cgroup `cpu.shares` and some heuristics around behavior when running with too few processors. This feature is enabled within a new `experimental` configuration block, with the intent of indicating the stability of the feature.
==COMMIT_MSG==

## Possible downsides?

Some initial attempts to implement this used the [containerd/cgroups](https://github.com/containerd/cgroups) library. Maybe I was doing something silly, but the API was a little bit awkward for our use case (reading cpu.shares), so I opted to parse the relevant files instead. This will probably need some changes if/when we need to add cgroup v2 support.


launchlib/processors.go
```diff
@@ -0,0 +1,148 @@
+// Copyright 2016 Palantir Technologies, Inc.
```
carterkozak: Perhaps the limit of my ability to review golang code ;-)
```suggestion
// Copyright 2023 Palantir Technologies, Inc.
```

launchlib/processors.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	cpuShareBytes, err := io.ReadAll(cpuSharesFile)
+	if err != nil {
+		return 0, errors.Wrapf(err, "unable to read cpu.shares")
+	}
+	cpuShares, err := strconv.Atoi(string(cpuShareBytes))
+	if err != nil {
+		return 0, errors.New("unable to convert cpu.shares value to expected type")
+	}
+	return uint(math.Max(2.0, math.Floor(float64(cpuShares/1024)))), nil
```
carterkozak: Possibly worth a link to https://github.com/palantir/go-java-launcher/issues/313 in a comment for context on the 2.0 minimum value.

carterkozak (commented):
Looks great, I really appreciate it @andybradshaw :-)

launchlib/processors.go
```diff
@@ -0,0 +1,148 @@
...
+
+type ProcessorCounter interface {
+	ProcessorCount() (uint, error)
+}
+
+var DefaultCGroupV1ProcessorCounter = CGroupV1ProcessorCounter{
+	fs: os.DirFS("/"),
+}
+
+type CGroupV1ProcessorCounter struct {
```
CRogers, line 47: I wonder if we should support cgroups v2 from the get go? Devs use docker-for-mac for dev-env tests, and docker-for-mac uses cgroups v2. It might be a little odd for it to behave differently on CI vs locally. The difference between the two is not that great, I had some code here which shows how cgroupsv2 works for memory at least: https://github.com/CRogers/gradle/commit/3fb0f7d6548aacf662d863b34a09b88c06a11e18.

launchlib/processors.go
```diff
@@ -0,0 +1,148 @@
...
+	fs fs.FS
+}
+
+func NewCGroupV1ProcessorCounter(filesystem fs.FS) ProcessorCounter {
+	return CGroupV1ProcessorCounter{fs: filesystem}
+}
+
+func (c CGroupV1ProcessorCounter) ProcessorCount() (uint, error) {
```
CRogers, line 55: When I was doing this for memory in gradle, there's a case where cgroups is installed but there is no limit for the current container/environment. An example of this is Circle machine executors (I wonder if PCloud is like this too?). Under cgroupsv1 you get [a really big value](https://github.com/CRogers/gradle/commit/3fb0f7d6548aacf662d863b34a09b88c06a11e18#diff-53102e5845be037fa8056d1330c50c4c74d69a5c9b9414f31c8c40fb2a75bb93R30-R32) (2^63) which means "no limit" but you can end up thinking there's more memory available than there is. Is this a problem as well for cpu shares/active processors implementation?

CRogers: The issue I want to avoid here is thinking we have 2^63 cpu shares, so tell java we have 2^63/1024 = 2^54 processors then witchcraft tries to make a thread pool with minimum 8 * 2^54 = 2^57 threads.

launchlib/processors.go
CRogers: Actually, in dev-env we restrict the active processors to 1 via a JVM arg by default, so I guess this doesn't really matter.

launchlib/processors.go
andybradshaw: As best as I can tell, groups get `1024` as their [cpu shares by default](https://kernel.googlesource.com/pub/scm/linux/kernel/git/glommer/memcg/+/cpu_stat/Documentation/cgroups/cpu.txt), so if someone has a very large value here, it would've had to be set manually. Maybe a safeguard here would be to detect the hosts number of processors? So something like:

```
max(2, min(runtime.NumCPUs(), floor(cpu.shares / 1024)))
```

launchlib/processors.go
andybradshaw: Updated... looks like the default godel license configuration handles the "year", so probably shouldn't have been [hardcoding 2016](https://github.com/palantir/go-java-launcher/blob/dd2531328daa91e4baa2eb85e9c43d7468cae7dc/godel/config/license-plugin.yml#L2).

launchlib/processors.go
andybradshaw: Looking at this now

launchlib/processors.go
carterkozak: I agree that cpu shares values shouldn't default to an extraordinarily high value (the jvm cgroups v1 implementation only appears to special case the 1024 default). Using `runtime.NumCPU` as a minimum sounds very reasonable (even taking precedence over the 2-processor minimum we assume based on shares, if we have confidence that the system has a single physical core).

The NumCPU implementation doesn't so anything exciting with cgroups, although I've only reviewed the linux implementation, the docs suggest that the intent is to describe the current process where `GOMAXPROCS` can be configured to hint go internals: https://github.com/golang/go/blob/f57f02fcd519a86683c47a444e0e24a086fea8e0/src/runtime/os_linux.go#L97-L122

launchlib/processors.go
andybradshaw: The range of valid `cpu.weight` values is quite a bit lower (1-10000) and I'm not entirely sure how this gets set/mapped in some situations. For example, it looks like [crun attempts to keep the same "shares" API](https://github.com/containers/crun/blob/main/crun.1.md#cpu-controller), and does a conversion behind the scenes. 

Looking through the kubernetes interactions now to see how requests end up getting translated, but want to know what we should treat as the value for "1 core"... should we use the cgroup v2 default weight (100)? should we use the translated cgroup v1 cpu shares weight (39)?

launchlib/processors.go
schlosna: https://github.com/uber-go/automaxprocs provides some similar functionality on the golang side for setting `GOMAXPROCS`

launchlib/processors.go
carterkozak: My preference would be consistency (avoiding special case values). Cgroups are a tool applied to a broad set of problems, where share/weight values are meant (in the context of pure cgroups) to provide relative weights as opposed to mapping to physical cores.
However, our use case targets container environments more specifically, where we have the additional context that the weights are based on millicore values. If container mode is used in go-java-launcher, we should assume a k8s (or similar) environment.

launchlib/processors.go
CRogers: fwiw I checked on an internal machine Circle executor and the cpu shares does appear to be 1024 and not some huge number (where as the memory limit is 2^63).

CRogers:
:+1:

launchlib/processors.go
carterkozak: @andybradshaw I'm happy merging this now, and handling cgroupsv2 in a separate PR. Whichever way you prefer.

carterkozak (approved):

PR #322, opened 2023-04-03
Refactor cgroup parsing to be less CPU module specific

andybradshaw:
## Before this PR

Previously, the cgroup parsing code had some bits specific to reading the cpu and cpu.shares information. We're going to need this for reading memory information as well, so should probably break it out into an interface.

### DIFF
--- a/launchlib/cgroup.go
+++ b/launchlib/cgroup.go
@@ -1,0 +1,111 @@
+package launchlib
+
+import (
+	"bufio"
+	"bytes"
+	"io"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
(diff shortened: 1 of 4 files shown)


launchlib/cgroup.go
```diff
@@ -0,0 +1,111 @@
...
+
+const (
+	selfCGroup    = "/proc/self/cgroup"
+	selfMountinfo = "/proc/self/mountinfo"
+)
+
+type CGroupName string
+
+type CGroupPather interface {
+	Path(name CGroupName) (string, error)
```
andybradshaw, line 23: Could see an argument for pushing the "value" reading here as well, so all of the io/fs path normalization stuff happens internally.

carterkozak (approved):

PR #323, opened 2023-04-03
Add dynamic heap percentage

andybradshaw:
## Before this PR

Currently, we set the RAM Percentage to 75% in all scenarios, which leads to a reasonable amount of waste in cases where we have large memory resource requests or limits. We should attempt to make this value dynamic based on the size of the cgroup that we're running in.

## After this PR

==COMMIT_MSG==
Add dynamic heap percentage system property
==COMMIT_MSG==

## Possible downsides?

Definitely need to play around with the function... have no idea what a good approach is 🙃


andybradshaw:
Probably going to do https://github.com/palantir/go-java-launcher/pull/324 first and start gathering data before adding anything around the scaling function

PR #329, opened 2023-04-25
Enable active processor count overriding by default

andybradshaw:
## After this PR

==COMMIT_MSG==
Enable active processor count overriding by default
==COMMIT_MSG==

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,8 +151,11 @@ All output from `go-java-launcher` itself, and from the launch of all processes 
 By default, when starting a java process inside a container (as indicated by the presence of ``CONTAINER`` env
 variable):
 
+1. `-XX:ActiveProcessorCount` will be set based on the discovered cgroup configurations and host information to a value
+   between 2 and the number of processors reported by the runtime. You can read more about the reasoning behind this
+   [here](https://github.com/palantir/go-java-launcher/issues/313).
 1. Args with prefix``-Xmx|-Xms`` in both static and custom jvm opts will be filtered out.
-2. If neither ``-XX:MaxRAMPercentage=`` nor ``-XX:InitialRAMPercentage=`` prefixes are present in either static or
+1. If neither ``-XX:MaxRAMPercentage=`` nor ``-XX:InitialRAMPercentage=`` prefixes are present in either static or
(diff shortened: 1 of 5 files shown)


launchlib/config.go
```diff
@@ -71,9 +71,7 @@ type CustomLauncherConfig struct {
 	DisableContainerSupport bool                       `yaml:"dangerousDisableContainerSupport"`
 }
 
-type ExperimentalLauncherConfig struct {
-	OverrideActiveProcessorCount bool `yaml:"overrideActiveProcessorCount"`
-}
+type ExperimentalLauncherConfig struct{}
```
carterkozak, line 74: on deserialization, do we allow unknown fields?

andybradshaw: Good question... I don't think we do strict deserialization, so should be safe to remove. If we were using `yaml.UnmarshlStrict` in the code snippet below, we'd run into failures:
https://github.com/palantir/go-java-launcher/blob/c4689c6d2d148db9aeafca7214ed6ee4d3d6b4a9/launchlib/config.go#L241

carterkozak: Great, thanks!

andybradshaw:
I want to double checking something real quick on a place where this is deployed... for some reason I remember seeing a higher number of processors the last time I looked at integration tests for this. just want to make sure we're not doing something silly and always returning `2` here 🙃

README.md
```diff
@@ -151,8 +151,11 @@ All output from `go-java-launcher` itself, and from the launch of all processes
 By default, when starting a java process inside a container (as indicated by the presence of ``CONTAINER`` env
 variable):
 
+1. `-XX:ActiveProcessorCount` will be set based on the discovered cgroup configurations and host information to a value
+   between 2 and the number of processors reported by the runtime. You can read more about the reasoning behind this
+   [here](https://github.com/palantir/go-java-launcher/issues/313).
```
carterkozak: Worth mentioning that if a `-XX:ActiveProcessorCount` value is already set, we don't modify it?

carterkozak (approved):

andybradshaw:
Confirmed internally that a larger number of cpu shares does lead to the "right" active processor arg getting set

PR #361, opened 2023-11-13
Add the `containerV2` flag to compute initial heap percentage using processor count, and disable setting the `ActiveProcessorCount` JVM option

mpritham:
## Before this PR
This is part of a larger effort in improving stability of Java services in Rubix.

### `-XX:ActiveProcessorCount`
k8s resource requests provide a minimum bound on resource usage. The `-XX:ActiveProcessorCount` JVM option acts as an upper bound on resource usage in several internal JDK features (e.g. GC threads, compiler threads). We currently set the active processor count based on resource requests. This has led to unpredictable performance under load.

### Setting initial heap size to be aware of the number of processors
From the above change, we're no longer limiting the number of processors available to the JVM. In the past, when we've removed limits on the number of processors (due to [JDK-8281181](https://bugs.openjdk.org/browse/JDK-8281181) being released in a critical patch update), we've encountered several OOM exceptions. This PR allocates 3mb per processor, by removing that much from the heap size. The rough 3mb figure was obtained by observation (cc: @carterkozak).

## After this PR
==COMMIT_MSG==
When the experimental `containerV2` is set, 1) compute the heap percentage as 75% of the heap minus 3mb per processor, with a minimum value of 50%, and 2) don't set the `-XX:ActiveProcessorCount` JVM option.
==COMMIT_MSG==

## Possible downsides?
This change will potentially waste available memory for services which would benefit from larger heap sizes. Also, the heap size heuristic is simple; a more sophisticated solution would allow for better heap utilization (see #323).

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,12 +165,11 @@ percentage of ``MaxRAM`` value, e.g. ``max-heap-size = MaxRAM * MaxRamPercentage
 If the experimental flag `containerV2` is set:
 1. The `-XX:ActiveProcessorCount` is unset, it will remain unset.
 1. Args with prefix``-Xmx|-Xms`` in both static and custom jvm opts will be filtered out. If neither 
- ``-XX:MaxRAMPercentage=`` nor ``-XX:InitialRAMPercentage=`` prefixes are present in either static or custom jvm opts,
- ``-Xmx|-Xms`` will both be set to be 75% of the heap size minus 3mb per processor, with a minimum value of 50% of the 
-  heap.
-1. If neither ``-XX:MaxRAMPercentage=`` nor ``-XX:InitialRAMPercentage=`` prefixes are present in either static or
-   custom jvm opts, both will be set to ``75.0`` (i.e. ``-XX:InitialRAMPercentage=75.0 -XX:MaxRAMPercentage=75.0 `` will
-   be appended after all the other jvm opts).
+ ``-XX:MaxRAMPercentage=`` nor ``-XX:InitialRAMPercentage=`` prefixes are present in either static or custom jvm opts:
+    - if we can obtain the cgroups memory limit ``-Xmx|-Xms`` will both be set to be 75% of the cgroups memory limit 
+     minus 3mb per processor, with a minimum value of 50% of the heap.
+    - if we cannot obtain the cgroups memory limit, both RAM percentage values will be set to ``75.0`` (i.e. 
+    ``-XX:InitialRAMPercentage=75.0 -XX:MaxRAMPercentage=75.0 `` will be appended after all the other jvm opts)
 
 ### Overriding default values
 
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ All output from `go-java-launcher` itself, and from the launch of all processes 
 
 ## Java heap and container support
 
-By default, when starting a java process inside a container (as indicated by the presence of ``CONTAINER`` env
+By _default_, when starting a java process inside a container (as indicated by the presence of ``CONTAINER`` env
 variable):
 
 1. If `-XX:ActiveProcessorCount` is unset, it will be set based on the discovered cgroup configurations and host
@@ -161,6 +161,16 @@ variable):
 
 This will cause the JVM 11+ to discover the ``MaxRAM`` value using Linux cgroups, and calculate the heap sizes as the specified
 percentage of ``MaxRAM`` value, e.g. ``max-heap-size = MaxRAM * MaxRamPercentage``.
+
+If the experimental flag `experimentalContainerV2` is set:
+1. The `-XX:ActiveProcessorCount` is unset, it will remain unset.
+1. Args with prefix``-Xmx|-Xms`` in both static and custom jvm opts will be filtered out. If neither 
+ ``-XX:MaxRAMPercentage=`` nor ``-XX:InitialRAMPercentage=`` prefixes are present in either static or custom jvm opts,
+ ``-Xmx|-Xms`` will both be set to be 75% of the heap size minus 3mb per processor, with a minimum value of 50% of the 
+  heap.
+1. If neither ``-XX:MaxRAMPercentage=`` nor ``-XX:InitialRAMPercentage=`` prefixes are present in either static or
+   custom jvm opts, both will be set to ``75.0`` (i.e. ``-XX:InitialRAMPercentage=75.0 -XX:MaxRAMPercentage=75.0 `` will
+   be appended after all the other jvm opts).
 
 ### Overriding default values
 
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ variable):
 This will cause the JVM 11+ to discover the ``MaxRAM`` value using Linux cgroups, and calculate the heap sizes as the specified
 percentage of ``MaxRAM`` value, e.g. ``max-heap-size = MaxRAM * MaxRamPercentage``.
 
-If the experimental flag `experimentalContainerV2` is set:
+If the experimental flag `containerV2` is set:
 1. The `-XX:ActiveProcessorCount` is unset, it will remain unset.
 1. Args with prefix``-Xmx|-Xms`` in both static and custom jvm opts will be filtered out. If neither 
  ``-XX:MaxRAMPercentage=`` nor ``-XX:InitialRAMPercentage=`` prefixes are present in either static or custom jvm opts,
--- a/integration_test/go_java_launcher_integration_test.go
+++ b/integration_test/go_java_launcher_integration_test.go
@@ -71,25 +71,25 @@ func TestMainMethodContainerSupportEnabled(t *testing.T) {
 			expectedJVMArgs: "-XX\\:InitialRAMPercentage=79.9 -XX\\:MaxRAMPercentage=80.9 -XX\\:ActiveProcessorCount=2",
 		},
 		{
-			name:               "using experimentalContainerV2 sets Xms and Xmx and does not set ActiveProcessorCount",
+			name:               "using containerV2 sets Xms and Xmx and does not set ActiveProcessorCount",
 			launcherCustom:     "testdata/launcher-custom-experimental-container-v2.yml",
 			expectedJVMArgs:    "",
 			expectedJVMArgKeys: []string{"-Xmx", "-Xms"},
 		},
 		{
-			name: "using experimentalContainerV2 with InitialRAMPercentage does not set Xms, Xmx, or " +
+			name: "using containerV2 with InitialRAMPercentage does not set Xms, Xmx, or " +
 				"ActiveProcessorCount",
 			launcherCustom:  "testdata/launcher-custom-experimental-container-v2-with-initial-ram-percentage.yml",
 			expectedJVMArgs: "-XX\\:InitialRAMPercentage=70.0",
 		},
 		{
-			name: "using experimentalContainerV2 with MaxRAMPercentage does not set Xms, Xmx, or " +
+			name: "using containerV2 with MaxRAMPercentage does not set Xms, Xmx, or " +
 				"ActiveProcessorCount",
 			launcherCustom:  "testdata/launcher-custom-experimental-container-v2-with-max-ram-percentage.yml",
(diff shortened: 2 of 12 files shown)


integration_test/testdata/launcher-custom-experimental-processor-aware-heap-pct-flag.yml
```diff
@@ -3,4 +3,4 @@ configVersion: 1
 jvmOpts:
   - '-Xmx1g'
 experimental:
-  overrideActiveProcessorCount: true
+  useProcessorAwareInitialHeapSize: true
```
mpritham: The `overrideActiveProcessorCount` was removed.

launchlib/launcher.go
```diff
@@ -311,13 +312,17 @@ func filterHeapSizeArgs(args []string) []string {
 	}
 
 	if !hasInitialRAMPercentage && !hasMaxRAMPercentage {
-		filtered = append(filtered, "-XX:InitialRAMPercentage=75.0")
-		filtered = append(filtered, "-XX:MaxRAMPercentage=75.0")
+		initialHeapPercentage, err := computeInitialHeapPercentage(customConfig)
+		if err != nil {
+			initialHeapPercentage = 75.0
+		}
+		filtered = append(filtered, fmt.Sprintf("-XX:InitialRAMPercentage=%.1f", initialHeapPercentage))
```
mpritham: I don't think additional precision beyond 1 decimal place will have much of an effect, but am open to changing this.

launchlib/memory.go
```diff
@@ -0,0 +1,71 @@
+// Copyright 2023 Palantir Technologies, Inc.
```
mpritham, line 1: Identical to the same file in #324

launchlib/memory_test.go
```diff
@@ -0,0 +1,95 @@
+// Copyright 2023 Palantir Technologies, Inc.
```
mpritham, line 1: Very similar (renamed the test) to the file in #324

mpritham:
Converting back to draft. I'd like to add a test.

launchlib/config.go
```diff
@@ -71,7 +71,10 @@ type CustomLauncherConfig struct {
 	DisableContainerSupport bool                       `yaml:"dangerousDisableContainerSupport"`
 }
 
-type ExperimentalLauncherConfig struct{}
+type ExperimentalLauncherConfig struct {
+	// Also disables setting the active processor count JVM argument.
+	UseProcessorAwareInitialHeapSize bool `yaml:"useProcessorAwareInitialHeapSize"`
```
carterkozak: Let's call this `experimentalContainerV2` instead of specifically how it works.

launchlib/launcher.go
carterkozak: Instead of computing specific percentage, let's set `-Xmx` and `-Xms` based on the desired heap size, since those args are better known/understood than the `Initial`/`MaxRAMPercentage`

mpritham: Updated. [It looks like the `Initial`/`MaxRAMPercentage` is used when `-Xmx`/`-Xms` is not set](https://github.com/openjdk/jdk/blob/master/src/hotspot/share/runtime/arguments.cpp#L1588-L1590), so I'm setting the flags inside the `if !hasInitialRAMPercentage && !hasMaxRAMPercentage` branch.

launchlib/launcher.go
```diff
@@ -376,3 +397,20 @@ func isMaxRAMPercentage(arg string) bool {
...
+		return 0, errors.Wrap(err, "failed to get cgroup processor count")
+	}
+	cgroupMemoryLimitInBytes, err := DefaultMemoryLimit.MemoryLimitInBytes()
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to get cgroup memory limit")
+	}
+	var heapLimit = float64(cgroupMemoryLimitInBytes)
+	var processorAdjustment = 3 * BytesInMebibyte * float64(cgroupProcessorCount)
+	var computedHeapSize = max(0.5*heapLimit, (0.75*heapLimit-processorAdjustment)/heapLimit)
+	return uint64(computedHeapSize), nil
```
mpritham: This conversion will truncate the fractional portion of the float.

launchlib/launcher.go
```diff
@@ -376,3 +397,20 @@ func isMaxRAMPercentage(arg string) bool {
...
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to get cgroup processor count")
+	}
+	cgroupMemoryLimitInBytes, err := DefaultMemoryLimit.MemoryLimitInBytes()
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to get cgroup memory limit")
+	}
+	var heapLimit = float64(cgroupMemoryLimitInBytes)
+	var processorAdjustment = 3 * BytesInMebibyte * float64(cgroupProcessorCount)
+	var computedHeapSize = max(0.5*heapLimit, (0.75*heapLimit-processorAdjustment)/heapLimit)
```
mpritham: This is wrong

integration_test/go_java_launcher_integration_test.go
```diff
@@ -69,6 +69,28 @@ func TestMainMethodContainerSupportEnabled(t *testing.T) {
...
+		{
+			name: "using experimentalContainerV2 with MaxRAMPercentage does not set Xms, Xmx, or " +
+				"ActiveProcessorCount",
+			launcherCustom:  "testdata/launcher-custom-experimental-container-v2-with-max-ram-percentage.yml",
+			expectedJVMArgs: "-XX\\\\:MaxRAMPercentage=70.0",
+		},
+		{
+			name:            "using experimentalContainerV2 does not use user-provided Xms or Xmx",
+			launcherCustom:  "testdata/launcher-custom-experimental-container-v2.yml",
+			expectedJVMArgs: "-Xms3107979264 -Xmx3107979264",
```
mpritham: What are your thoughts on just checking if the `Xms` and `Xmx` values are set in the test rather than a specific figure?

launchlib/config.go
```diff
@@ -71,7 +71,9 @@ type CustomLauncherConfig struct {
 	DisableContainerSupport bool                       `yaml:"dangerousDisableContainerSupport"`
 }
 
-type ExperimentalLauncherConfig struct{}
+type ExperimentalLauncherConfig struct {
+	ExperimentalContainerV2 bool `yaml:"experimentalContainerV2"`
```
carterkozak: Perhaps we call this `containerV2` since the parent section is called `experimental` (sorry I originally suggested including experimental in the prop name!)

mpritham: Changed. Thought you were being extra explicit that this was experimental! 😄

launchlib/launcher.go
```diff
@@ -276,12 +281,21 @@ func delim(str string) string {
...
-func createJvmOpts(combinedJvmOpts []string, customConfig *CustomLauncherConfig, logger io.WriteCloser) []string {
+func createJvmOpts(combinedJvmOpts []string, customConfig *CustomLauncherConfig, logger io.WriteCloser) ([]string, error) {
 	if isEnvVarSet("CONTAINER") && !customConfig.DisableContainerSupport && !hasMaxRAMOverride(combinedJvmOpts) {
 		_, _ = fmt.Fprintln(logger, "Container support enabled")
-		combinedJvmOpts = filterHeapSizeArgs(combinedJvmOpts)
-		combinedJvmOpts = ensureActiveProcessorCount(combinedJvmOpts, logger)
-		return combinedJvmOpts
+		if customConfig.Experimental.ContainerV2 {
+			jvmOptsWithUpdatedHeapSizeArgs, err := filterHeapSizeArgsV2(combinedJvmOpts)
+			if err != nil {
```
carterkozak, line 286: Let's leave a comment for ourselves here describing the fallback to percentage-based heap sizing. Worth noting that it's not the ideal solution because it doesn't take into account the per-processor memory offset, but has the upside of already supporting platforms that are supported by Java, so the trade-off makes sense for now.

launchlib/launcher.go
```diff
@@ -376,3 +418,12 @@ func isMaxRAMPercentage(arg string) bool {
 func isInitialRAMPercentage(arg string) bool {
 	return strings.HasPrefix(arg, "-XX:InitialRAMPercentage=")
 }
+
+// ComputeJVMHeapSizeInBytes If the experimental `ContainerV2` is set, compute the heap size to be 75% of
+// the heap minus 3mb per processor, with a minimum value of 50% of the heap.
+func ComputeJVMHeapSizeInBytes(hostProcessorCount int, cgroupMemoryLimitInBytes uint64) uint64 {
+	var heapLimit = float64(cgroupMemoryLimitInBytes)
```
carterkozak: terminology nit: Let's call this the memory limit, we'll never be able to launch a jvm with a heap very close to this value

launchlib/launcher.go
```diff
@@ -276,12 +281,21 @@ func delim(str string) string {
...
+			jvmOptsWithUpdatedHeapSizeArgs, err := filterHeapSizeArgsV2(combinedJvmOpts)
+			if err != nil {
+				combinedJvmOpts = filterHeapSizeArgs(combinedJvmOpts)
+			} else {
+				combinedJvmOpts = jvmOptsWithUpdatedHeapSizeArgs
+			}
+		} else {
+			combinedJvmOpts = filterHeapSizeArgs(combinedJvmOpts)
+		}
+		combinedJvmOpts = ensureActiveProcessorCount(customConfig, combinedJvmOpts, logger)
```
carterkozak: I believe this statement is only meaningful in the `else` block in on 295 when `customConfig.Experimental.ContainerV2` is false. Should we move it there?

mpritham: Sounds good. I was checking for the experimental flag inside that method and not adding the active processor count option if the flag was set, but I can move it to 295 and remove the flag check.

Saves us the needless for-loop over the arguments.

README.md
```diff
@@ -162,6 +162,16 @@ variable):
 This will cause the JVM 11+ to discover the ``MaxRAM`` value using Linux cgroups, and calculate the heap sizes as the specified
 percentage of ``MaxRAM`` value, e.g. ``max-heap-size = MaxRAM * MaxRamPercentage``.
 
+If the experimental flag `containerV2` is set:
+1. The `-XX:ActiveProcessorCount` is unset, it will remain unset.
+1. Args with prefix``-Xmx|-Xms`` in both static and custom jvm opts will be filtered out. If neither 
+ ``-XX:MaxRAMPercentage=`` nor ``-XX:InitialRAMPercentage=`` prefixes are present in either static or custom jvm opts,
+ ``-Xmx|-Xms`` will both be set to be 75% of the heap size minus 3mb per processor, with a minimum value of 50% of the 
+  heap.
+1. If neither ``-XX:MaxRAMPercentage=`` nor ``-XX:InitialRAMPercentage=`` prefixes are present in either static or
```
carterkozak: I think we need to clarify that this case only applies when cgroup memory limits aren't detected

launchlib/launcher.go
```diff
@@ -276,12 +281,24 @@ func delim(str string) string {
 	return fmt.Sprintf("%s%s%s", TemplateDelimsOpen, str, TemplateDelimsClose)
 }
 
-func createJvmOpts(combinedJvmOpts []string, customConfig *CustomLauncherConfig, logger io.WriteCloser) []string {
+func createJvmOpts(combinedJvmOpts []string, customConfig *CustomLauncherConfig, logger io.WriteCloser) ([]string, error) {
```
carterkozak: I don't think the `error` component of the result is used anymore

README.md
carterkozak: It may also be fair to exclude the fallback info from the documentation as it shouldn't be reachable in rubix

launchlib/launcher.go
mpritham: 🤦 yep.

README.md
mpritham: I updated it to specify that the fallback is used when we're not able to get the cgroup memory limit.

carterkozak (approved):
Thanks!

mpritham:
👍

PR #369, opened 2024-01-09
Enable `containersV2` by default

mpritham:
## Before this PR
In #361 we introduced the `containersV2` flag. After observing improvements across several stacks, we'd like to enable the functionality guarded by the flag as the default behavior. We'd like to keep the option open for devs to disable the behavior if needed, by setting the `containersV2` flag to `false`.

## After this PR
==COMMIT_MSG==
Enable `containersV2` by default.
==COMMIT_MSG==

## Possible downsides?
N/A.

### DIFF
--- a/integration_test/go_java_launcher_integration_test.go
+++ b/integration_test/go_java_launcher_integration_test.go
@@ -144,7 +144,7 @@ func TestMainMethodWithoutCustomConfig(t *testing.T) {
 }
 
 func TestMainMethodContainerWithoutCustomConfig(t *testing.T) {
-	output := testContainerSupportEnabled(t, "foo", "-XX\\:InitialRAMPercentage=75.0 -XX\\:MaxRAMPercentage=75.0 -XX\\:ActiveProcessorCount=2", []string{})
+	output := testContainerSupportEnabled(t, "foo", "", []string{"-Xmx", "-Xms"})
 	assert.Regexp(t, `Failed to read custom config file, assuming no custom config: foo`, output)
 }
 
--- a/integration_test/testdata/launcher-custom-dangerous-disable-container-support.yml
+++ b/integration_test/testdata/launcher-custom-dangerous-disable-container-support.yml
@@ -3,3 +3,5 @@ configVersion: 1
 jvmOpts:
   - '-Xmx1g'
 dangerousDisableContainerSupport: true
+experimental:
+  containerV2: false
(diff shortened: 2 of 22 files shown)


integration_test/go_java_launcher_integration_test.go
```diff
@@ -144,7 +144,7 @@ func TestMainMethodWithoutCustomConfig(t *testing.T) {
 }
 
 func TestMainMethodContainerWithoutCustomConfig(t *testing.T) {
-	output := testContainerSupportEnabled(t, "foo", "-XX\\:InitialRAMPercentage=75.0 -XX\\:MaxRAMPercentage=75.0 -XX\\:ActiveProcessorCount=2", []string{})
+	output := testContainerSupportEnabled(t, "foo", "", []string{"-Xmx", "-Xms"})
```
mpritham, line 147: If there is a no custom config, or there is a failure to read the custom config, we use the new default behavior, hence this change.

integration_test/testdata/launcher-custom-dangerous-disable-container-support.yml
```diff
@@ -3,3 +3,5 @@ configVersion: 1
 jvmOpts:
   - '-Xmx1g'
 dangerousDisableContainerSupport: true
+experimental:
```
mpritham, line 6: Setting the flag to `false` to retain tests for old behavior for now.

launchlib/config.go
```diff
@@ -267,6 +267,10 @@ func parseCustomConfig(yamlString []byte) (PrimaryCustomLauncherConfig, error) {
 				"subProcess config %s", name)
 		}
 	}
+	if config.Experimental.ContainerV2 == nil {
+		var trueVal = true
+		config.Experimental.ContainerV2 = &trueVal
```
mpritham, line 272: I've seen this pattern elsewhere, but it would be nice if I could make this one line.

andybradshaw: I don't think there's anything in the standard library for it :( but there are some pretty small deps for doing something like [this](https://go.dev/play/p/vwkCLlBqQSm):
```golang
package main

import (
    "fmt"
    "k8s.io/utils/ptr"
)

func main() {
    a := ptr.To(false)
    fmt.Printf("%T %v", a, *a)
}

```

mpritham: I see. I'm leaning towards not adding the dep since we use this in one place in the implementation (and also once in the tests).

andybradshaw (approved):

carterkozak:
Thanks @mpritham and @andybradshaw! 👍

PR #466, opened 2024-09-13
Remove the experimental `ContainerV2` flag

mpritham:
## Before this PR
In #369, we enabled the use of the logic behind the `containersV2` flag by default. Internally, there are no usages of `containerV2: false`. We can remove the flag, and the fallback behavior now. 

## After this PR
==COMMIT_MSG==
Remove the experimental `ContainerV2` flag
==COMMIT_MSG==

### DIFF
--- a/launchlib/config.go
+++ b/launchlib/config.go
@@ -72,7 +72,6 @@ type CustomLauncherConfig struct {
 }
 
 type ExperimentalLauncherConfig struct {
-	ContainerV2 *bool `yaml:"containerV2"`
 }
 
 type PrimaryCustomLauncherConfig struct {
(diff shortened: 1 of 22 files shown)


launchlib/config.go
```diff
@@ -72,7 +72,6 @@ type CustomLauncherConfig struct {
 }
 
 type ExperimentalLauncherConfig struct {
```
mpritham, line 74: Arguably, we can remove this. Opting to keep it to make it easier to add experimental feature flags in the future.

carterkozak: 👍

carterkozak (approved):

PR #548, opened 2025-05-05
Handle CGroupV2

parker89:
## Before this PR

CI has been broken for a couple months due to [CircleCI upgrading to CGroupV2](https://discuss.circleci.com/t/docker-executor-infrastructure-upgrade/52282). Additionally we have been using CGroupV2 more commonly so it's necessary to update our logic and tests to handle CGroupV2. 

I noticed that processors.go was not used so I deleted it and its tests. 

## After this PR

==COMMIT_MSG==
Handle CGroupV2 in addition to CGroupV1
==COMMIT_MSG==


launchlib/memory.go
```diff
@@ -26,26 +25,36 @@ import (
...
 type CGroupMemoryLimit struct {
 	pather CGroupPather
 	fs     fs.FS
 }
 
-func NewCGroupMemoryLimit(filesystem fs.FS) MemoryLimit {
+func NewCGroupMemoryLimit(filesystem fs.FS) (MemoryLimit, error) {
+	isCGroupV2, err := IsCGroupV2(filesystem)
+	if err != nil {
+		return CGroupMemoryLimit{}, errors.Wrap(err, "failed to determine cgroup version")
```
mpritham: I see in other places, when there's an error we're propagating, we usually use `nil` as the return type. Thoughts on doing that here?

Don't have much go experience. Your call.

launchlib/memory.go
```diff
@@ -54,18 +63,58 @@ func (c CGroupMemoryLimit) MemoryLimitInBytes() (uint64, error) {
...
+		return 0, errors.New("unable to convert memory limit value to expected type")
 	}
 	return uint64(memLimit), nil
 }
+
+func (c CGroupMemoryLimit) readCGroupV1MemoryLimit(memoryCGroupPath string) ([]byte, error) {
+	cgroupV1MemLimitFilepath := filepath.Join(memoryCGroupPath, cgroupV1MemLimitName)
+	cgroupV1MemLimitFile, err := c.fs.Open(convertToFSPath(cgroupV1MemLimitFilepath))
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to open memory.limit_in_bytes at expected location: %s", cgroupV1MemLimitFile)
```
mpritham:
```suggestion
		return nil, errors.Wrapf(err, "unable to open memory.limit_in_bytes at expected location: %s", cgroupV1MemLimitFilepath)
```

When there's an error opening the file, I suspect the return value is `nil` ([maybe src](https://github.com/golang/go/blob/dad4f39971d89b56224d1eb44121305b1c0ef711/src/os/root.go#L327))? Thoughts on using the filepath instead?

launchlib/memory.go
```diff
@@ -54,18 +63,58 @@ func (c CGroupMemoryLimit) MemoryLimitInBytes() (uint64, error) {
...
+	memLimitBytes, err := io.ReadAll(cgroupV1MemLimitFile)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to read memory.limit_in_bytes")
+	}
+	return memLimitBytes, nil
+}
+
+func (c CGroupMemoryLimit) readCGroupV2MemoryLimit(memoryCGroupPath string) ([]byte, error) {
+	cgroupV2MemLimitFilepath := filepath.Join(memoryCGroupPath, cgroupV2MemLimitName)
+	cgroupV2MemLimitFile, err := c.fs.Open(convertToFSPath(cgroupV2MemLimitFilepath))
```
mpritham, line 117: From [these docs](https://pkg.go.dev/io/fs#FS), I think we need to call `cgroupV2MemLimitFile.close()` at some point?

launchlib/memory.go
```diff
@@ -54,18 +63,58 @@ func (c CGroupMemoryLimit) MemoryLimitInBytes() (uint64, error) {
 		return 0, errors.Wrap(err, "failed to get memory cgroup path")
 	}
 
-	memLimitFilepath := filepath.Join(memoryCGroupPath, memLimitName)
-	memLimitFile, err := c.fs.Open(convertToFSPath(memLimitFilepath))
+	isCGroupV2, err := IsCGroupV2(c.fs)
```
mpritham: We're checking `isCGroupV2` in `NewCGroupMemoryLimit` as well. What are your thoughts on extending `CGroupMemoryLimit` with a field for `version` or `boolean isCgroupV2`? Can avoid a double-check I think.

mpritham (dismissed):
Approving for an RC!

parker89:
Verified the RC produces the behavior expected

mpritham (approved):
Thanks!

PR #557, opened 2025-05-20
Add heapPercentage configuration

parker89:
## Before this PR

Overriding ``-XX:MaxRAMPercentage=`` and ``-XX:InitialRAMPercentage=`` to the same value does not guarantee that the heap will not shrink. This is different behavior than setting  ``-Xmx`` and ``-Xms``  to the same value.

The default go-java-launcher behavior adjusts the heap size based on the number of processors, resulting in ``-Xmx`` and ``-Xms`` both being set to the same value. The new `heapPercentage` configuration allows services to override the default 75% value while also preventing heap shrinking. 

Note the new `heapPercentage` does not apply the per-processor adjustment. It is assumed anyone overriding this can account for this themselves. 

More detail: 
``-Xms`` sets the minimum heap size. There is no equivalent for `RAMPercentage` style JVM options (note that while `MinRAMPercentage` does exist, it has a confusing name and instead controls the max RAM percentage at memory sizes under ~200MB). 

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/launchlib/config.go
+++ b/launchlib/config.go
@@ -69,6 +69,7 @@ type CustomLauncherConfig struct {
 	Env                     map[string]string          `yaml:"env"`
 	Experimental            ExperimentalLauncherConfig `yaml:"experimental"`
 	DisableContainerSupport bool                       `yaml:"dangerousDisableContainerSupport"`
+	HeapPercentage          *float64                   `yaml:"heapPercentage" validate:"gt=0,lte=100"`
 }
 
 type ExperimentalLauncherConfig struct {
--- a/launchlib/config_test.go
+++ b/launchlib/config_test.go
@@ -299,6 +299,7 @@ jvmOpts:
   - jvmOpt1
   - jvmOpt2
 dangerousDisableContainerSupport: true
+heapPercentage: 75
 `,
 			want: PrimaryCustomLauncherConfig{
 				VersionedConfig: VersionedConfig{
@@ -310,6 +311,7 @@ dangerousDisableContainerSupport: true
 					},
 					JvmOpts:                 []string{"jvmOpt1", "jvmOpt2"},
 					DisableContainerSupport: true,
+					HeapPercentage:          &[]float64{75}[0],
 					Experimental:            ExperimentalLauncherConfig{},
 				},
 			},
--- a/launchlib/config_test.go
+++ b/launchlib/config_test.go
@@ -312,7 +312,7 @@ heapPercentage: 75
 					},
 					JvmOpts:                 []string{"jvmOpt1", "jvmOpt2"},
 					DisableContainerSupport: true,
-					HeapPercentage:          &[]float64{75}[0],
+					HeapPercentage:          toPointer(75.0),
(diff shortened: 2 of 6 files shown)


launchlib/config.go
```diff
@@ -69,6 +69,7 @@ type CustomLauncherConfig struct {
 	Env                     map[string]string          `yaml:"env"`
 	Experimental            ExperimentalLauncherConfig `yaml:"experimental"`
 	DisableContainerSupport bool                       `yaml:"dangerousDisableContainerSupport"`
+	HeapPercentage          *float64                   `yaml:"heapPercentage" validate:"gt=0,lte=100"`
```
mpritham, line 72: Does the `validate` block make `go` throw an error when it's out of this range? Should we add a test for that (maybe it's obvious/not needed)?

launchlib/launcher.go
```diff
@@ -388,14 +392,18 @@ func isInitialRAMPercentage(arg string) bool {
...
+// ComputeJVMHeapSizeInBytes By default, compute the heap size to be 75% of the heap minus 3mb per processor, with a minimum value
+// of 50% of the heap. If heapPercentage is provided, use that percentage instead with no processor adjustment.
+func ComputeJVMHeapSizeInBytes(hostProcessorCount int, cgroupMemoryLimitInBytes uint64, heapPercentage *float64) (uint64, error) {
 	if cgroupMemoryLimitInBytes > 1_000_000*BytesInMebibyte {
 		return 0, errors.New("cgroups memory limit is unusually high. Not computing JVM heap size")
 	}
 	var memoryLimit = float64(cgroupMemoryLimitInBytes)
 	var processorAdjustment = 3 * BytesInMebibyte * float64(hostProcessorCount)
+
+	if heapPercentage != nil {
```
mpritham, line 403: nit: thoughts on moving this up 1 line: we don't use `processorAdjustment`

launchlib/config_test.go
```diff
@@ -299,6 +300,7 @@ jvmOpts:
   - jvmOpt1
   - jvmOpt2
 dangerousDisableContainerSupport: true
+heapPercentage: 75
```
mpritham, line 303: should we update the README, or is this something you don't want to make broadly visible right now?

kunalrkak (commented):
overall looks sane to me! will defer to pritham's review

launchlib/config.go
parker89: Yeah it throws an error: https://pkg.go.dev/gopkg.in/validator.v2

No need to test this IMO

launchlib/launcher.go
parker89: Done

launchlib/config_test.go
parker89: I've updated the README to include this - I expect people to use it. I think this will be strictly better than manually overriding `MaxRAMPercentage` and `InitialRAMPercentage`

mpritham (approved):
lgtm, thanks!

PR #571, opened 2025-06-26
Add support for native image execution mode

mpritham:
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/launchlib/config.go
+++ b/launchlib/config.go
@@ -73,6 +73,11 @@ type CustomLauncherConfig struct {
 }
 
 type ExperimentalLauncherConfig struct {
+	// Acceptable values are "native" and "jvm". If omitted, the default value is "jvm".
+	ExecutionMode ExecutionMode `yaml:"executionMode,omitempty"`
+	// NativeImageArguments specifies additional arguments that will be passed to the executable when running in native execution mode.
+	NativeImageArguments      []string `yaml:"nativeImageArguments"`
+	NativeImageExecutablePath string   `yaml:"nativeImageExecutablePath"`
 }
 
 type PrimaryCustomLauncherConfig struct {
@@ -98,6 +103,13 @@ var allowedLauncherConfigs = AllowedLauncherConfigValues{
 		"grafana-server": {},
 		"envoy":          {}},
 }
+
+type ExecutionMode string
+
+const (
+	ExecutionModeNative ExecutionMode = "native"
+	ExecutionModeJvm    ExecutionMode = "jvm"
+)
 
 func GetConfigsFromFiles(
 	staticConfigFile string, customConfigFile string, stdout io.Writer) (
@@ -267,6 +279,9 @@ func parseCustomConfig(yamlString []byte) (PrimaryCustomLauncherConfig, error) {
 				"subProcess config %s", name)
 		}
 	}
+	if err := validateExecutionMode(config.Experimental.ExecutionMode); err != nil {
+		return PrimaryCustomLauncherConfig{}, err
+	}
 	return config, nil
 }
 
@@ -323,4 +338,14 @@ func convertMap(intMap map[int]struct{}) map[string]struct{} {
 		stringMap[strconv.Itoa(k)] = v
 	}
 	return stringMap
+}
+
+func validateExecutionMode(mode ExecutionMode) error {
+	if mode == "" {
+		return nil
+	}
+	if mode != ExecutionModeNative && mode != ExecutionModeJvm {
+		return fmt.Errorf("experimental.executionMode must be one of {'%s', '%s'} (default '%s') found '%s'", ExecutionModeNative, ExecutionModeJvm, ExecutionModeJvm, mode)
+	}
+	return nil
 }
--- a/launchlib/config.go
+++ b/launchlib/config.go
@@ -106,6 +106,8 @@ var allowedLauncherConfigs = AllowedLauncherConfigValues{
 
 type ExecutionMode string
 
+// ExecutionModeNative indicates that the application should be launched as a GraalVM native image executable. ExecutionModeJvm
+// indicates that the application should be launched as a JVM application.
 const (
 	ExecutionModeNative ExecutionMode = "native"
 	ExecutionModeJvm    ExecutionMode = "jvm"
--- a/launchlib/config_test.go
+++ b/launchlib/config_test.go
@@ -342,6 +342,32 @@ env:
 				},
 			},
 		},
+		{
+			name: "nonexistent key",
+			data: `
+configType: executable
+configVersion: 1
+nonexistentKey: value
+env:
+  SOME_ENV_VAR: /etc/profile
+  OTHER_ENV_VAR: /etc/redhat-release
+`,
+			want: PrimaryCustomLauncherConfig{
+				VersionedConfig: VersionedConfig{
+					Version: 1,
+				},
+				CustomLauncherConfig: CustomLauncherConfig{
+					TypedConfig: TypedConfig{
+						Type: "executable",
+					},
+					Env: map[string]string{
+						"SOME_ENV_VAR":  "/etc/profile",
+						"OTHER_ENV_VAR": "/etc/redhat-release",
+					},
+					Experimental: ExperimentalLauncherConfig{},
+				},
+			},
+		},
 	} {
 		got, _ := parseCustomConfig([]byte(currCase.data))
 		assert.Equal(t, currCase.want, got, "Case %d: %s", i, currCase.name)
--- a/launchlib/config_test.go
+++ b/launchlib/config_test.go
@@ -368,6 +368,41 @@ env:
 				},
 			},
 		},
+		{
+			name: "executionMode is parsed",
+			data: `
+configType: java
+configVersion: 1
+experimental:
+  executionMode: native
+  nativeImageExecutablePath: /service/bin/my-application-native
+  nativeImageArguments:
+    - arg1
+    - arg2
+env:
+  SOME_ENV_VAR: /etc/profile
+  OTHER_ENV_VAR: /etc/redhat-release
+`,
+			want: PrimaryCustomLauncherConfig{
+				VersionedConfig: VersionedConfig{
+					Version: 1,
+				},
+				CustomLauncherConfig: CustomLauncherConfig{
+					TypedConfig: TypedConfig{
+						Type: "java",
+					},
+					Env: map[string]string{
+						"SOME_ENV_VAR":  "/etc/profile",
+						"OTHER_ENV_VAR": "/etc/redhat-release",
+					},
+					Experimental: ExperimentalLauncherConfig{
+						ExecutionMode:             ExecutionModeNative,
+						NativeImageExecutablePath: "/service/bin/my-application-native",
+						NativeImageArguments:      []string{"arg1", "arg2"},
+					},
+				},
+			},
+		},
 	} {
 		got, _ := parseCustomConfig([]byte(currCase.data))
 		assert.Equal(t, currCase.want, got, "Case %d: %s", i, currCase.name)
--- a/launchlib/launcher.go
+++ b/launchlib/launcher.go
@@ -86,7 +86,19 @@ func compileCmdFromConfig(
 	var executable string
 	var executableErr error
 
-	if staticConfig.Type == "java" {
+	// Handle experimental nativeImage execution mode before standard processing
+	if customConfig.Experimental.ExecutionMode == ExecutionModeNative {
+		executable, executableErr = verifyPathIsSafeForExec(customConfig.Experimental.NativeImageExecutablePath)
+		if executableErr != nil {
+			return nil, executableErr
+		}
+		args = append(args, executable) // 0th argument is the command itself
+		// Add default system properties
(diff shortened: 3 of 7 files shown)


launchlib/config_test.go
```diff
@@ -322,6 +322,32 @@ heapPercentage: 75
...
+					Env: map[string]string{
+						"SOME_ENV_VAR":  "/etc/profile",
+						"OTHER_ENV_VAR": "/etc/redhat-release",
+					},
+					Experimental: ExperimentalLauncherConfig{},
+				},
+			},
+		},
+		{
+			name: "nonexistent key",
```
aldexis, line 346: Can we also add a test around parsing the execution mode?

launchlib/launcher.go
```diff
@@ -86,7 +86,33 @@ func compileCmdFromConfig(
 	var executable string
 	var executableErr error
 
-	if staticConfig.Type == "java" {
+	// Handle experimental nativeImage execution mode before standard processing
+	if customConfig.Experimental.ExecutionMode == ExecutionModeNative {
```
aldexis: Thoughts on nesting this case under `if staticConfig.Type == "java"`? (or otherwise check that it's always java if execution mode is native?) I'm not sure what `staticConfig.Type == "executable"` is though

launchlib/launcher.go
```diff
@@ -86,7 +86,33 @@ func compileCmdFromConfig(
...
+		executable, executableErr = verifyPathIsSafeForExec(customConfig.Experimental.NativeImageExecutablePath)
+		if executableErr != nil {
+			return nil, executableErr
+		}
+		args = append(args, executable) // 0th argument is the command itself
+		// Add default system properties
+		args = append(args, "-Djava.io.tmpdir=var/data/tmp", "-Dsun.net.inetaddr.ttl=20")
```
aldexis: I don't see these being set in other places - why are we specifically setting this for executable?

launchlib/config.go
```diff
@@ -99,6 +104,13 @@ var allowedLauncherConfigs = AllowedLauncherConfigValues{
 		"envoy":          {}},
 }
 
+type ExecutionMode string
+
+const (
+	ExecutionModeNative ExecutionMode = "native"
+	ExecutionModeJvm    ExecutionMode = "jvm"
```
aldexis, line 113: nit: Maybe add a comment explaining what this is meant for?

launchlib/launcher.go
```diff
@@ -86,7 +86,33 @@ func compileCmdFromConfig(
...
+		for _, nativeArg := range customConfig.Experimental.NativeImageArguments {
+			args = append(args, nativeArg)
+		}
```
aldexis: Are we not passing any of the jvm opts? I guess it's not jvm, but aren't some options the same? (I don't know if it makes sense to share the configs, but am wondering how usable this will be if people need to maintain two sets of options)

launchlib/launcher_test.go
```diff
@@ -148,3 +149,73 @@ func TestMkdirChecksDirectorySyntax(t *testing.T) {
...
+		},
+	}
+
+	cgroups := map[string]string{}
+	createLogger := func() (io.WriteCloser, error) { return &NoopClosingWriter{io.Discard}, nil }
+
+	cmd, err := compileCmdFromConfig(&staticCfg, &customCfg, &cgroups, createLogger)
+	require.NoError(t, err)
+
+	assert.Contains(t, cmd.Args, "-XX:MaxRAMPercentage=60.0")
```
aldexis: vnit: Use the value from `heap` here, rather than hardcode `60.0` a second time

aldexis (commented):
Overall lgtm, but probably worth @parker89 to have a look as well

aldexis (dismissed):
Approving to allow cutting a release candidate

launchlib/launcher.go
parker89: Yes there's some overlap, but I think it will be too difficult to maintain a list of which opts apply to native vs JVM. 

Better to just make it explicit

launchlib/launcher.go
```diff
@@ -86,7 +86,33 @@ func compileCmdFromConfig(
...
+		// If heapPercentage is specified, ensure MaxRAMPercentage flag is present
+		if customConfig.HeapPercentage != nil {
+			hasFlag := false
+			for _, arg := range customConfig.Experimental.NativeImageArguments {
+				if strings.HasPrefix(arg, "-XX:MaxRAMPercentage=") {
```
parker89: The native flag is actually `-XX:MaximumHeapSizePercent` 
https://www.graalvm.org/latest/reference-manual/native-image/optimizations-and-performance/MemoryManagement/

launchlib/launcher.go
parker89: And `-XX:+IgnoreUnrecognizedVMOptions` isn't valid in Graal native, so if we were to fail to strip out an unrecognized native option, the native image would fail to start up.

launchlib/launcher.go
parker89: I agree - makes the code path a little more clear that `executionMode` is a sub-option of `staticConfig.Type = "java"`

launchlib/launcher.go
parker89: Looks like it comes from here: https://github.com/palantir/sls-packaging/blob/develop/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfig.java#L69

Could we instead bake these options into the native image at build time via our gradle plugin? That would be most equivalent to how we do it for JVM programs

launchlib/launcher.go
```diff
@@ -86,7 +86,33 @@ func compileCmdFromConfig(
...
+		if executableErr != nil {
+			return nil, executableErr
+		}
+		args = append(args, executable) // 0th argument is the command itself
+		// Add default system properties
+		args = append(args, "-Djava.io.tmpdir=var/data/tmp", "-Dsun.net.inetaddr.ttl=20")
+		for _, nativeArg := range customConfig.Experimental.NativeImageArguments {
+			args = append(args, nativeArg)
+		}
+		// If heapPercentage is specified, ensure MaxRAMPercentage flag is present
```
parker89: nit: I'd say something like: 

// If `MaximumHeapSizePercent` is not present in `NativeImageArguments`, use `heapPercentage`.

launchlib/launcher.go
mpritham: > Could we instead bake these options into the native image at build time via our gradle plugin?

yeah this makes sense.

launchlib/launcher_test.go
parker89: You're saying instead of: 
`	assert.Contains(t, cmd.Args, "-XX:MaxRAMPercentage=60.0")`
it should be something like:
`	assert.Contains(t, cmd.Args, fmt.Sprintf("-XX:MaxRAMPercentage=%.1f", heap))`
Am I understanding that correctly? 

That seems to somewhat defeat the purpose of the test since we'd just be copying over the code we're trying to test. So if the actual string format command should have been `%.1n` or something (making it up obviously) then our test would miss it. Simpler and more robust to just hardcode both the string and the float IMO

launchlib/launcher.go
```diff
@@ -86,7 +86,33 @@ func compileCmdFromConfig(
...
+		if customConfig.HeapPercentage != nil {
+			hasFlag := false
+			for _, arg := range customConfig.Experimental.NativeImageArguments {
+				if strings.HasPrefix(arg, "-XX:MaxRAMPercentage=") {
+					hasFlag = true
+					break
+				}
+			}
+			if !hasFlag {
+				maxRAMFlag := fmt.Sprintf("-XX:MaxRAMPercentage=%.1f", *customConfig.HeapPercentage)
```
parker89: Can we make this `%.2f` instead? Since we also round `heapPercentage` to the hundredth decimal in some of our internal tooling.

bjlaub (dismissed):

launchlib/launcher.go
```diff
@@ -367,6 +405,18 @@ func filterHeapSizeArgsV2(args []string, heapPercentage *float64) ([]string, err
...
+	var nativeArgs []string
+	for _, arg := range jvmOpts {
+		for _, allowed := range AllowedNativeImageJVMOptions {
+			if strings.HasPrefix(arg, allowed) {
+				nativeArgs = append(nativeArgs, arg)
+			}
+		}
+	}
+	return nativeArgs
+}
```
aldexis: Thoughts on putting this and the whitelist into a separate file? (I imagine the whitelist will grow over time and it might be beneficial for it to live separately, and will make it easier to realize that it's only based on prefix matching, which will require some changes if we want certain options)

launchlib/launcher.go
```diff
@@ -87,30 +97,58 @@ func compileCmdFromConfig(
...
+		if customConfig.Experimental.ExecutionMode == ExecutionModeNative {
+			executable, executableErr = verifyPathIsSafeForExec(customConfig.Experimental.NativeImageExecutablePath)
+			if executableErr != nil {
+				return nil, executableErr
+			}
+			args = append(args, executable) // 0th argument is the command itself
+
+			args = append(args, getNativeArgsFromJVMOpts(combinedJvmOpts)...)
+
+			args = append(args, customConfig.Experimental.NativeImageArguments...)
```
aldexis: From the existing code, I deduct that the JVM will use the last value for an option if it's present multiple times. Have we confirmed that the native image would behave in the same way? (might also be worth a quick comment explaining that we're porting over some specific jvm options from the defaults/custom, but also prioritizing any option that is explicitly set for native images)

launchlib/launcher.go
parker89: Moved some native logic over to native.go

launchlib/launcher.go
parker89: That's correct. Added a comment clarifying this.

launchlib/launcher.go
```diff
@@ -109,24 +99,11 @@ func compileCmdFromConfig(
 			}
 			args = append(args, executable) // 0th argument is the command itself
 
+			// Filter out any heap-related options that are not supported by the current mode (container mode enabled/disabled), and adds -XX:MaximumHeapSizePercent using heapPercentage if applicable
```
aldexis: This seems to imply that we're _only_ filtering heap-related options, but that's not all we're doing, right? We're also applying a whitelist to the jvm opts

launchlib/native.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+	}
+	return nativeArgs
+}
+
+// getNativeArgs filters out any heap-related options that are not supported by the current mode (container mode enabled/disabled), and adds -XX:MaximumHeapSizePercent using heapPercentage if applicable
+func getNativeArgs(nativeArgs []string, customConfig *CustomLauncherConfig) []string {
+	var filteredArgs []string
+
+	if isEnvVarSet("CONTAINER") && !customConfig.DisableContainerSupport {
```
aldexis, line 61: Why are the options different between running in a container vs not? 🤔 

And should we be applying this heap more globally to all args, including those coming from the jvm opts? (not that we _currently_ include any heap setting in the whitelist, but this would seem more resilient if we applied to `combinedJvmOpts` + `customConfig.Experimental.NativeImageArguments` rather than just `customConfig.Experimental.NativeImageArguments` as we're currently doing)

aldexis (dismissed):
Approving to unblock RC

launchlib/native.go
parker89: That's how it works for [standard JVM currently](https://github.com/palantir/go-java-launcher/blob/develop/launchlib/launcher.go#L281)

[Some docs on container mode](https://github.com/palantir/go-java-launcher?tab=readme-ov-file#java-heap-and-container-support)

parker89: I don't think `AllowedNativeImageJVMOptions` should ever contain any heap size opts - it's pretty clearly not the intended purpose. We have a comment there stating the list is adapted from the sls-packaging opts - which also wouldn't ever include anything heap related.

launchlib/launcher.go
parker89: Good catch - moved the comment down a line to make it clear which line it applies to.

aldexis (dismissed):
Re-approving for RC

launchlib/native.go
```diff
@@ -63,7 +64,8 @@ func getNativeArgs(nativeArgs []string, customConfig *CustomLauncherConfig) []st
 
 		// If MaximumHeapSizePercent is not present in NativeImageArguments and container mode is not disabled, use heapPercentage for -XX:MaximumHeapSizePercent
 		if !hasMaximumHeapSizePercentOverride(nativeArgs) && customConfig.HeapPercentage != nil {
-			maxHeapSizeFlag := fmt.Sprintf("-XX:MaximumHeapSizePercent=%.2f", *customConfig.HeapPercentage)
+			// MaximumHeapSizePercent must be an integer
```
aldexis, line 67:
```suggestion
			// MaximumHeapSizePercent must be an integer in native images
```

aldexis (dismissed):
Approving for RC

bjlaub (approved):
thanks!

PR #669, opened 2026-03-06
Add allowHeapShrink experimental flag

parker89:
## Summary
- Adds `allowHeapShrink` boolean to `experimental` config. When `true`, it only affects the cgroup-based heap sizing path: the launcher omits the generated `-Xms` flag and strips `-XX:+AlwaysPreTouch`, allowing the JVM heap to start small and grow/shrink dynamically instead of being pinned to max.
- Only takes effect in container mode (`CONTAINER` env var set, `dangerousDisableContainerSupport: false`, no `-XX:MaxRAM` override) when cgroup-based heap sizing is used.
- Defaults to `false`, preserving current behavior (min == max).

## Changes
- `config.go`: New `AllowHeapShrink bool` field on `ExperimentalLauncherConfig`
- `launcher.go`: Thread flag through the cgroup-based heap sizing path, conditionally skip `-Xms` append and strip `-XX:+AlwaysPreTouch`
- `launcher.go`: Move cgroup memory lookup out of `filterHeapSizeArgsV2` so the helper can be tested with an explicit memory limit
- `config_test.go`: YAML parsing test for `allowHeapShrink: true`
- `launcher_test.go`: Deterministic unit tests for V2 `allowHeapShrink` behavior using an explicit cgroup memory limit, including `AlwaysPreTouch` preservation/removal, RAM-percentage override preservation, and oversized-memory error handling

## Test plan
- [x] `./godelw test` passes (launchlib tests green; integration test failure is pre-existing `syscall.SIGPOLL` on macOS)
- [x] `./godelw format` clean
- [ ] CI passes on Linux


launchlib/launcher.go
```diff
@@ -300,19 +300,18 @@ func delim(str string) string {
...
 			// When we fail to get the memory limit from the cgroups files, fallback to using percentage-based heap
 			// sizing. While this method doesn't take into account the per-processor memory offset, it is supported
 			// by all platforms using Java.
 			// Also, when the memory limit is unusually high (defined to be over 1TB), we revert to the
 			// percentage-based heap sizing. This is to handle the edge case where the cgroups memory limit is set
 			// to be an arbitrary large value.
 			combinedJvmOpts = filterHeapSizeArgs(combinedJvmOpts, customConfig.HeapPercentage)
-		} else {
-			combinedJvmOpts = jvmOptsWithUpdatedHeapSizeArgs
+			return combinedJvmOpts
```
mpritham: Thoughts on logging the error here? e.g. if the `allowHeapShrink` flag is set but the service is not able to read the cgroup memory limit, the flag won't be respected.

mpritham: Also maybe the `memory limit is unusually high. Not computing JVM heap size` message is useful for debugging?

mpritham (dismissed):
LGTM

launchlib/launcher.go
parker89: Good call. Split the error / check out into 2 different if blocks so I can log the reason.

mpritham (dismissed):

mpritham (approved):

PR #768, opened 2026-09-02
enable single node ZGC testing

occasional-handbook:
## Before this PR

we cannot test ZGC on a single node, it is all-or-nothing. Instead, we could allow a single node to use ZGC, and compare performance more directly.

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/launchlib/config.go
+++ b/launchlib/config.go
@@ -96,6 +96,9 @@ type ExperimentalLauncherConfig struct {
 	// Deprecated: use the top-level allowHeapShrink field on CustomLauncherConfig instead. This field is retained
 	// for backwards compatibility with existing overrides that set it under experimental.
 	AllowHeapShrink bool `yaml:"allowHeapShrink,omitempty"`
+	// ZGCCanaryHostnameSuffix replaces the SLS Packaging GC profile options on the host whose name has the
+	// configured suffix.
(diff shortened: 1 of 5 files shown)


parker89 (dismissed):
+1 for RC

launchlib/config.go
```diff
@@ -96,6 +96,9 @@ type ExperimentalLauncherConfig struct {
 	// Deprecated: use the top-level allowHeapShrink field on CustomLauncherConfig instead. This field is retained
 	// for backwards compatibility with existing overrides that set it under experimental.
 	AllowHeapShrink bool `yaml:"allowHeapShrink,omitempty"`
+	// ZGCCanaryHostnameSuffix replaces the SLS Packaging GC profile options on the host whose name has the
+	// configured suffix.
+	ZGCCanaryHostnameSuffix string `yaml:"zgcCanaryHostnameSuffix,omitempty"`
```
ash211, line 101: if we were to release this beyond RC, I think we should change this to be more broad and flexible.

For example, an option where you can more generally override config on canary node0, for some fraction of nodes.  This way you can canary settings other than GC (like string interning settings `-XX:+UseStringDeduplication`), and also canary different fractions of nodes (instead of just one node: node0).

